PR #47, opened 2015-01-19
Add <Context> path attribute as a parameter for jira class

poikilotherm:
This PR will add template code to be able to add a context path for the JIRA Tomcat instance. 

I need this in companion with my reverse proxy, as I have a context path in my apache in which I add the reverse proxy configuration.

I added some validations in init.pp, too. (For $contextpath, $proxy and $db).

### DIFF
--- a/README.md
+++ b/README.md
@@ -298,6 +298,10 @@ If the jira service is managed outside of puppet the stop_jira paramater can be 
 
 Defaults to {}, See examples on how to use.
 
+#####`$contextpath = ""`
+
+Defaults to an empty string (""). Will add a path to the Tomcat Server Context.
+
 ####Tomcat parameters####
 
 #####`$tomcatPort`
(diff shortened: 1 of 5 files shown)


mkrakowitzer:
Thank you :) Would you consider taking a stab at adding some tests. 

I recently added the context path to the stash module, so you could use that as an example:
https://github.com/mkrakowitzer/puppet-stash/blob/master/spec/classes/stash_config_spec.rb

poikilotherm:
Yeah sure, I'll write a test...
I just thought there are currently none and I did not want to get in someones way... ;-)

poikilotherm:
I eat my words...
Of course there were already spec tests, just been dazzled.

I added the test for the context path.

Slightly off-topic: would you like me to refactor the init.pp and params.pp? I have some ideas for that... I'll gladly open an issue to sum them up.

mkrakowitzer:
Certainly open to it, I would say open an issue and we take it from there.

Thanks for your contributions. Oh and please consider rating the Module on the forge if you have the time :)

Ill get this merged ASAP, just want to run it through my beaker tests, altough its a no brainer it will pass.

PR #51, opened 2015-01-21
Turn on the AJP connector based on a hiera or puppet-variable lookup

peculater:
I tried to make it similar to the way that proxy is set up, such that you can add additional parameters to the Connector object.


mkrakowitzer:
Thank you. Would you consider adding some tests to: spec/classes/jira_config_spec.rb

Its not a requirement, but perhaps also add validate_hash($ajp) to init.pp

PR #58, opened 2015-02-10
Add a notify and subscribe options to the jira service

paulgeringer:
Use case is when something like JDK is updated underneath jira, you'd want the service to restart, with:
```puppet
subscribe => Package['jdk']
```
I tested separately, services work fine with something like:
 ```puppet 
notify => undef
```
in the service definition.

### DIFF
--- a/README.md
+++ b/README.md
@@ -310,6 +310,14 @@ Manage the JIRA service, defaults to 'running'
 
 Defaults to 'true'
 
+#####`$service_subscribe`
+
+Restart the jira service in response to this subscription
+
+#####`$service_notify`
+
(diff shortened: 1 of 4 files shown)


mkrakowitzer:
Thank you, would you consider adding some tests to spec/classes/jira_service_spec.rb?

paulgeringer:
In addition to the tests already added? I'm not sure what else to test there.

mkrakowitzer:
ok fair enough.

mkrakowitzer:
Thanks for your contribution. Please consider rating the jira module on puppet forge if you have some free time :)

PR #59, opened 2015-02-10
Added in some extra specs as requested

paulgeringer:
Following up from [here](https://github.com/brycejohnson/puppet-jira/pull/58#issuecomment-73761920), some extra specs. Turns out rspec-puppet is pretty awesome.

Thanks to @zachdaniel for help figuring out the extra package resource.


mkrakowitzer:
Thank you so much :+1:

Issue #54, opened 2015-01-30
Add tests for Microsoft SQL


mkrakowitzer:
Add tests for Microsoft SQL

mkrakowitzer:
Thanks!

PR #60, opened 2015-03-05
Add MS SQL Server tests

smsearcy:
Added tests for Microsoft SQL Server, per Issue #54.

I based it off the existing tests but was unable to run the test suite via rake due a protocol error unrelated to what I added (tried it without my modification).  This was the first time I've written or run spec tests so I think it is related to my setup, not necessarily the tests.

Also my apologies, I wasn't sure how to connect this to the existing issue.


mkrakowitzer:
Thanks for this! The first test is always the hardest :)

PR #67, opened 2015-04-01
Provide capability to set Tomcat's bind address

dougneal:
Add a new parameter, jira::tomcatAddress, which is passed through to the Tomcat's Connector configuration. Useful for environments where Jira is deployed behind a reverse proxy such as Nginx.

The parameter defaults to undef, and if left unset, the Connector configuration will not include the address= parameter (i.e. existing installations will be unchanged)

A new rspec test has been provided to test this functionality.

### DIFF
--- a/README.md
+++ b/README.md
@@ -335,6 +335,10 @@ Defaults to an empty string (""). Will add a path to the Tomcat Server Context.
 
 ####Tomcat parameters####
 
+#####`$tomcatAddress`
+
+IP address to listen on. Defaults to all addresses.
+
 #####`$tomcatPort`
 
 Port to listen on, defaults to 8080,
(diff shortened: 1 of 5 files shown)


mkrakowitzer:
I think we would need to update the jira::facts class

The uri is currently 127.0.0.1, which is fine, but if we specify the tomcatAddress, it should rather use that value?

dougneal:
Agreed - good catch.

poikilotherm:
This merge leads to some problems. If one does not specify a tomcatAddress, there will be a address="" in the <Connector>, which in turn lead to a binding of tcp ::ffff:127.0.0.01:8080 socket.

I think this is implied by the template incorrectly checking for the undef var. Will open a PR with the fix. Maybe better add a spec test for that, too.

mkrakowitzer:
Agreed, Thanks for assisting with a fix.

PR #69, opened 2015-04-13
Update Atlassian download URL

fgeek:


mkrakowitzer:
I don't understand the reason for changing the URL. The existing url seems to redirect correctly.
Mine redirects to: my.atlassian.com/software/jira/downloads/binary/atlassian-jira-X.X.X.tar.gz

fgeek:
Ok. I did not know about the redirect.

PR #74, opened 2015-04-24
added new parameter disable_notifications in relation to setenv.sh

f3rr:
in some cases its needed to comment in DISABLE_NOTIFICATIONS in setenv.sh. this pull request will add this function to this module.

### DIFF
--- a/spec/classes/jira_config_spec.rb
+++ b/spec/classes/jira_config_spec.rb
@@ -13,7 +13,8 @@ describe 'jira::config' do
             :version  => '6.3.4a',
             :javahome => '/opt/java',
           }}
-          it { should contain_file('/opt/jira/atlassian-jira-6.3.4a-standalone/bin/setenv.sh')}
+          it { should contain_file('/opt/jira/atlassian-jira-6.3.4a-standalone/bin/setenv.sh')
+            .with_content(/#DISABLE_NOTIFICATIONS=/) }
           it { should contain_file('/opt/jira/atlassian-jira-6.3.4a-standalone/bin/user.sh')}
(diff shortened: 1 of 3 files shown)


mkrakowitzer:
Thanks f3rr, would you consider adding a test?

f3rr:
Hey Mkrakowitzer, Tests added

spec/classes/jira_config_spec.rb
```diff
@@ -171,6 +172,15 @@
           it { should contain_file('/opt/jira/atlassian-jira-6.3.4a-standalone/conf/context.xml')
             .with_content(/<Resource name = "testdb"\n        auth = "Container"\n    \/>/) }
         end
+
+        context 'disable notifications' do
+          let(:params) {{
+            :version  => '6.3.4a',
+            :javahome => '/opt/java',
```
mkrakowitzer: Here we should specify the 'disable_notifications => true',

spec/classes/jira_config_spec.rb
```diff
@@ -171,6 +172,15 @@
...
+          }}
+          it { should contain_file('/opt/jira/atlassian-jira-6.3.4a-standalone/bin/setenv.sh')
+            .with_content(Regexp.new('^#DISABLE_NOTIFICATIONS=')) }
```
mkrakowitzer: This regex will need to be updated. Something like /^DISABLE_NOTIFICATIONS=/. 

You don't need to use Regexp.new, You can just use /match/.

spec/classes/jira_config_spec.rb
f3rr: sorry, my fault. Its fixed now. I hope its ok now.

mkrakowitzer:
:+1: thanks!

PR #78, opened 2015-05-28
Add examples to puppet-jira

mkrakowitzer:


igalic:
@mkrakowitzer this commit uses an email that's not in your github profile

examples/jira_mysql_install.pp
```diff
@@ -0,0 +1,24 @@
...
+
+  class { '::jira':
+    javahome    => '/opt/java/latest',
+    db          => 'mysql',
+    dbport      => '3306',
+    dbdriver    => 'com.mysql.jdbc.Driver',
+    dbtype      => 'mysql',
+  } ->
+
+  class { '::jira::facts': }
```
ffrank: I deeply resent the antipattern of declaring classes with the resource like syntax when no parameters are being passed. Does the `jira::facts` class really depend on the `jira` class? Otherwise I'd prefer to drop the `->` and use `include`.

Generally, is there a style guide that addresses this question? The [official docs](https://docs.puppetlabs.com/puppet/latest/reference/lang_classes.html#include-like-vs-resource-like) have a discussion at least, but no clear recommendation I think.

mkrakowitzer: To be honest for me its just semantics, I really don’t mind either way. I think include is more correct, so I'll update the example. The dependency is not required, just a bad habit where I tend to want to run things in the exact same order every time.

ffrank: Oh, that *is* a bad habit. Graph complexity is no laughing matter.

Cool, I'm a huge `include` fan (as you can probably tell). The comment pertains to some other examples as well.

nibalizer: I would support/use a lint plugin to formalize this.

igalic: :+1:

ffrank: I can't say [no](https://github.com/ffrank/puppetlint-classdeclerations) to you guys.

PR #80, opened 2015-06-11
Forces jira user to be local

thedonkdonk:
Because the module tries to modify the user, the module will fail if the user is not local. This will force the user be local. I can't imagine a use case where you would want the user to not be local and have this module modify the jira user. It's pretty much this or take out jira user management.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -29,7 +29,7 @@ class jira::install {
     managehome       => true,
     uid              => $jira::uid,
     gid              => $jira::gid,
-
+    forcelocal       => true,
   }
 
   if ! defined(File[$jira::installdir]) {


mkrakowitzer:
I think its certainly possible for someone to want manage the jira account in another NSS other than locally. It probably should be configurable?

Adding the feature to remove the jira user management is also welcome, this was recently added to the stash module: https://github.com/puppet-community/puppet-stash/pull/71

thedonkdonk:
I considered that. Didn't see anyone else have a pending issue on it yet, so I figured the likelihood was low but you are right. The flexibility should be there. I'll get a new pull request done soon.

igalic:
imo, if you want your jira user to be … more special, you should manage it yourself, rather than add onto this module.

Issue #76, opened 2015-05-15
Add SSL Support


Conzar:
Can you please add SSL [support](https://confluence.atlassian.com/display/JIRA/Running+JIRA+over+SSL+or+HTTPS#RunningJIRAoverSSLorHTTPS-commandline)?

stack72:
Do you think it is possible to use something like Nginx to SSL Proxy jira?

Conzar:
This seems to work without having to modify the jira configuration at all.  So maybe its not worth adding since apache can handle it?
{code}
  class { 'apache':
    mpm_module    => 'prefork',
    default_vhost => false,
    require       => Postgresql::Server::Db ['jira'],
  }

  class { 'apache::mod::php':
  }

  class { 'apache::mod::rewrite':
  }

  class { 'apache::mod::wsgi':
  }

  class { 'apache::mod::ssl':
  }
  apache::vhost { 'jira non-ssl':
    servername => 'jira',
    port       => '80',
    docroot    => '/var/www',
    rewrites   => [
      {
        comment      => 'redirect to https',
        rewrite_cond => ['%{HTTPS} off'],
        rewrite_rule => ['(.*) https://%{HTTP_HOST}:443%{REQUEST_URI}'],
      },
    ],
  }
  apache::vhost { 'jira ssl':
    servername          => 'jira',
    port                => '443',
    docroot             => '/var/www',
    default_vhost       => true,
    ssl                 => true,
    ssl_cert            => '/etc/ssl/certs/ssl-cert-snakeoil.pem',
    ssl_key             => '/etc/ssl/private/ssl-cert-snakeoil.key',
    proxy_preserve_host => 'On',
    proxy_pass => [
      { 'path' => '/', 'url' => 'http://localhost:8080/' },
    ],
    require => Apache::Vhost [ 'jira non-ssl']
  }
{code}

stack72:
@Conzar that's great news. I think the documentation for the module can just be updated to show your example code. I'm not a massive fan of complicating systems and SSL Termination works a lot of the time for me :)

Conzar:
Well, not so good news.  It seemed to work on a fresh install of Jira.  But it didn't work on an existing install of jira.  Essentially, it was getting a proxy timeout error.  Does anything need to change on the server.xml?  

Here is the main error:
```
[...]

PR #81, opened 2015-06-22
Add SSL Support #76

mkrakowitzer:
Add native ssl support

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -64,7 +64,6 @@ class jira (
   $mysql_connector_install = '/opt/MySQL-connector',
   $mysql_connector_URL     = 'http://cdn.mysql.com/Downloads/Connector-J',
 
-
   # Configure database settings if you are pooling connections
   $enable_connection_pooling = false,
   $poolMinSize               = 20,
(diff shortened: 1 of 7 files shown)


manifests/init.pp
```diff
@@ -132,6 +137,9 @@
   validate_re($contextpath, ['^$', '^/.*'])
   validate_hash($resources)
   validate_hash($ajp)
+  validate_bool($tomcatNativeSsl)
+  validate_absolute_path($tomcatKeystoreFile)
+  validate_re($tomcatKeystoreType, '^(JKS|JCEKS|PKCS12)$')
```
ffrank: Oh cool :)

Which remindes me, is there a canonical way of validating port numbers?

mkrakowitzer:
@ffrank updated with all your suggestions. If you are happy, please merge it.

Accidentally closed the previous pr and I can not reopen it as I force pushed my branch after the rebase, sorry :(

manifests/init.pp
mkrakowitzer: Not that i know of, seems something like below might work for between 1024 to 65535 and will not accept any 0000.
```
validate_re($port, '^(102[4-9]|10[3-9]\d|1[1-9]\d{2}|[2-9]\d{3}|[1-5]\d{4}|6[0-4]\d{3}|65[0-4]\d{2}|655[0-2]\d|6553[0-5])$')
```
From here: http://stackoverflow.com/questions/25450318/how-to-limit-regex-validation-for-port-number-validation-1024-to-65535

PR #92, opened 2015-09-17
Fix mysql connector containment

adamcrews:
When using the mysql_connector download, the file is often times downloaded after the service is started.  This fix uses the contain function to make sure the mysql_connector class is contained in the install class, thus fixing dependancies.

It also updates the travis tests and the metadata.json to use the minimum versions of puppet that have the contain class.  This means that anyone still using Puppet Enterprise < 3.2 will not be able to use this module anymore, but that version has been long unsupported by Puppet Labs so this should not be an issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,17 +5,17 @@ sudo: false
 matrix:
   include:
   - rvm: 1.9.3
-    env: PUPPET_GEM_VERSION="~> 3.3.0"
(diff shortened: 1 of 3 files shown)


adamcrews:
Hmm.. it looks like travis didnt see the PR.  The build does pass on my branch though: https://travis-ci.org/adamcrews/puppet-jira/builds/80906053

PR #94, opened 2015-10-07
7.0.0 - Support and Puppet 4.0

ffquintella:
This code runs on puppet 4.0 and Puppet Enterprise and has support for Jira 7.0.0

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -32,7 +32,7 @@ class jira::install {
 
   }
 
-  if ! defined(File[$jira::installdir]) {
+  if (!defined(File[$jira::installdir])) {
     file { $jira::installdir:
       ensure => 'directory',
       owner  => $jira::user,
@@ -43,7 +43,7 @@ class jira::install {
   $file = "atlassian-${jira::product}-${jira::version}.${jira::format}"
   if $jira::staging_or_deploy == 'staging' {
 
-    require staging
+    class{'staging':}
 
     if ! defined(File[$jira::webappdir]) {
       file { $jira::webappdir:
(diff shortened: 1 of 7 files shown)


manifests/install.pp
```diff
@@ -103,11 +113,11 @@
 
   if $jira::db == 'mysql' and $jira::mysql_connector_manage {
     if $jira::staging_or_deploy == 'staging' {
-      class { '::jira::mysql_connector':
+      class { 'jira::mysql_connector':
         require => Staging::Extract[$file],
       }
     } elsif $jira::staging_or_deploy == 'deploy' {
-      class { '::jira::mysql_connector':
+      class { 'jira::mysql_connector':
```
bastelfreak: why don't you want to use the top level scope?

manifests/install.pp
```diff
@@ -40,10 +40,20 @@
     }
   }
 
-  $file = "atlassian-${jira::product}-${jira::version}.${jira::format}"
+  # https://www.atlassian.com/software/jira/downloads/binary/atlassian-jira-software-7.0.0-jira-7.0.0-x64.bin
+
+  $arr_version = split($jira::version, '.')
+
+  $major = 0 + $arr_version[0]
```
bastelfreak: Why you add 0? An interesting method to cast something to an int?

manifests/install.pp
ffquintella: Not needed  and it creates problems with puppet 4.0 and lint

manifests/install.pp
ffquintella: This is a way to convert the string into a number. But this isn't  the last code. 

Int the end I just used $jira::version[0] 

Let me see why it came to you

ffquintella:
Sorry the last code had problems on the merge to the master this one should be the latest version

ffquintella:
This pull is wrong I missed up the merge.I will redo it

PR #97, opened 2015-10-09
Service status is not working when using init scripts

jlcox1970:
Turns out that the check for the status of the jira server was always returning a state of 'stopped'

This was due to the jps task always having an exit status of 1 even thou the jira service was running.
 have chaged this to simply look for a process owened by the jira user

### DIFF
--- a/templates/jira.initscript.erb
+++ b/templates/jira.initscript.erb
@@ -62,24 +62,18 @@ function start() {
 
 function status() {
 
-  STATUS=$( ps aux | grep "[c]atalina.base=$CATALINA_HOME" | wc -l )
-  if [ $STATUS -gt 0 ];then
-    RETVAL=`ps -ef |grep $SERVICE |grep -v grep |grep atlassian-jira |grep ^<%= scope.lookupvar('jira::user')%> |wc -l`
-    if [ $RETVAL -ne 0 ];then
-      echo "$SERVICE is running"
-      return 0
-    else
-      echo "$SERVICE is stopped"
-      return 1
-    fi
-  else
-    echo "$SERVICE is stopped"
-    return 1
+STATUS=$( pgrep -u <%= scope.lookupvar('jira::user')%> -f atlassian-jira ) || STATUS=0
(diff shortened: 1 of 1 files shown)


templates/jira.initscript.erb
```diff
@@ -64,14 +64,13 @@ function status() {
 
   STATUS=$( ps aux | grep "[c]atalina.base=$CATALINA_HOME" | wc -l )
   if [ $STATUS -gt 0 ];then
-    ps -ef |grep $SERVICE |grep -v grep |awk '{ print $2 }' | <%= scope.lookupvar('jira::javahome') %>/bin/jps |grep -v Jps |grep -v grep > /dev/null
-    RETVAL=$?
-    if [ $RETVAL -eq 0 ];then
+    RETVAL=`ps -ef |grep $SERVICE |grep -v grep |grep atlassian-jira |grep ^<%= scope.lookupvar('jira::user')%> |wc -l`
```
igalic:
```
pgrep -u <%= scope.lookupvar('jira::user')%>  -f atlassian-jira
```

templates/jira.initscript.erb
```diff
@@ -64,14 +64,13 @@ function status() {
 
   STATUS=$( ps aux | grep "[c]atalina.base=$CATALINA_HOME" | wc -l )
```
igalic: this is pretty awful, too…

templates/jira.initscript.erb
igalic: come to think of it, this whole set of cascaded ifs could be that line…

jyaworski:
Hello:

What's the status of this PR?

bastelfreak:
HI @jlcox1970, could you take a look at the improvements @igalic mentioned?

jlcox1970:
@bastelfreak I have made some more changes in line of what @igalic was suggesting
My testing shows that it is working as before but the code I think is much cleaner

bastelfreak:
thanks!

PR #98, opened 2015-10-12
Paramaterizing more tomcat settings

jacobmw:


mkrakowitzer:
Would you consider adding tests for the new parameters? Some examples here:
https://github.com/puppet-community/puppet-jira/blob/master/spec/classes/jira_config_spec.rb

I am OK with the camel case parameters as its consistent with the previous parameters. Just keep in mind that we will rename them all with the next major version bump which will happen soon, Likely when Jira 7.0 support is added.

jacobmw:
I'll go ahead and add tests for these new paramaters.  I do imagine that this entire module is going to need to change significantly to support Jira 7.0

mkrakowitzer:
@jacobmw yeah I think you are right, I think it will be best to take the same approach as I am with stash, v1.x of this module will support up to the latest release of Jirav6. v2 of this module will support an upgrade from last release of v6 to v7.X.X

mkrakowitzer:
Thanks, can you squash this into one commit.

igalic:
thanks @jacobmw!

PR #108, opened 2015-12-11
Adding configuration option to enable or disable websudo.

ronaldvanrij:
Adding configuration option to enable or disable websudo.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -44,7 +44,7 @@ class jira (
   $gid          = undef,
   $shell        = '/bin/true',
 
-  $no_websudo   = false,
+  $enable_secure_admin_sessions = true,
 
   # Database Settings
   $db                      = 'postgresql',
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -44,6 +44,8 @@ class jira (
   $gid          = undef,
   $shell        = '/bin/true',
 
+  $no_websudo   = false,
+
   # Database Settings
   $db                      = 'postgresql',
   $dbuser                  = 'jiraadm',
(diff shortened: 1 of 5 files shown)


igalic:
@ronaldvanrij something called web***sudo*** should definitely be documented.

manifests/init.pp
```diff
@@ -44,6 +44,8 @@
   $gid          = undef,
   $shell        = '/bin/true',
 
+  $no_websudo   = false,
```
igalic: this naming is absolutely confusing.

ronaldvanrij: Agreed on the confusing parameter, unfortunately this reversed approach has been decided on by Atlassian, see the configuration item "jira.websudo.is.disabled". Reversing their decision by naming this parameter "websudo" may cause confusion as well. Nevertheless I'll update the parameter to a positive approach to controlling this feature.

juniorsysadmin: Mirroring the the ugliness of the original variable is probably the better option: eg. ```$jira_websudo_is_disabled```

igalic: software is sadness, made by people for other people.

JimPanic: The Jira documentation says it is this: "Disabling Secure Administrator Sessions"; why not make it `$disable_secure_admin_sessions = false`?

JimPanic: (`websudo` *is* a bad name)

igalic: because *disable*_secure_admin_sessions, again, is confusing ;)

JimPanic: Uhm. Okay. Then `$enable_secure_admin_sessions = true`. Both better than `anything_websudo`

igalic: :+1:

ronaldvanrij:
I've renamed the parameter to "enable_secure_admin_sessions". Added tests and updated Readme.MD.

JimPanic:
<3

igalic:
thank you so much @ronaldvanrij!

PR #110, opened 2015-12-11
Change Default Download URL and use HTTPS

bramwelt:
The download URL appears to have moved as a lookup on the current URL
returns two HTTP 301 responses and one HTTP 302.

This patch replaced the current download URL with the redirected one and switches the protocol to use HTTPS.
```
$ curl -sIL http://www.atlassian.com/software/jira/downloads/binary/atlassian-jira-6.4.12.tar.gz | egrep '(HTTP|Location)'
HTTP/1.1 301 Moved Permanently
Location: https://www.atlassian.com/software/jira/downloads/binary/atlassian-jira-6.4.12.tar.gz
HTTP/1.1 301 Moved Permanently
Location: https://my.atlassian.com/software/jira/downloads/binary/atlassian-jira-6.4.12.tar.gz
HTTP/1.1 302 Found
Location: https://downloads.atlassian.com/software/jira/downloads/atlassian-jira-6.4.12.tar.gz
HTTP/1.1 200 OK
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -299,7 +299,7 @@ defaults to ''
 #####`$downloadURL`
 
 The URL used to download the JIRA installation file.
-Defaults to 'http://www.atlassian.com/software/jira/downloads/binary/'
+Defaults to 'https://downloads.atlassian.com/software/jira/downloads/'
 
 #####`$staging_or_deploy`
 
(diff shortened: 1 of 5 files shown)


igalic:
thank you @bramwelt!

Issue #112, opened 2016-01-03
Unable to provide own configuration options in jira-config.properties


oysteing:
After commit e2f929f820b03af7d81c508714e298d533841bec, I'm not able to set custom configuration in jira-config.properties anymore, as the file is now controlled by puppet-jira.

There are no options for providing own properties, and the only reason for managing this file was to disable WebSudo.

A number of solution would be acceptable to most users:
- Don't manage the file unless the default setting is overridden
- Manage the file with Augeas
- Allow users to provide a map of key/value pairs to be appended to the file

igalic:
@oysteing all of those sound like sensible options.
i think i'd prefer the third.
would you like to take a stab at providing a patch?

oysteing:
Sure. I've created pull request #113.

PR #99, opened 2015-10-16
Systemd support for debian 8 jessie

whitewolf712:
On Debian Jessie can use systemd units instead Init.d-scripts.
When I try use old init.d scheme with jira, i have error because puppet can't use 'service' command for control init.d based services on Jessie.

I try to use jira systemd unit without any changes and its work fine.

PR #113, opened 2016-01-07
Attempt at solving #112 by providing key/value hash

oysteing:


spec/classes/jira_config_spec.rb
```diff
@@ -351,6 +351,20 @@
...
+          let(:params) {{
+            :version  => '6.3.4a',
+            :javahome => '/opt/java',
+            :jiraConfigProperties => {
+              'ops.bar.group.size.opsbar-transitions' => '4',
+            }
+          }}
+          it { should contain_file('/home/jira/jira-config.properties')
+            .with_content(/jira.websudo.is.disabled = false/) }
+            .with_content(/ops.bar.group.size.opsbar-transitions = 4/) }
```
rnelson0: I don't think you need these trailing `}`'s

igalic: yeah, not like this is a sensible language like lisp :P

oysteing:
Embarassing. I blame it on the Lisp gnome sneaking those braces in.

What do you think of the request in general?
* Hash solution vs other solutions?
* The new parameter?
* Camel case? There is a mix of camel case and snake case, I chose according to my preference, but feel free to instruct.
* Merging the enable_secure_admin_sessions setting into the hash?

igalic:
camelCase is discouraged in the [style guide](https://docs.puppetlabs.com/guides/style_guide.html#variable-format)

igalic:
if we remove `enable_secure_admin_sessions` we're breaking compatibility, if we merge we have to consider both names

oysteing:
I've changed to snake_case. If there are no more comments, I consider myself done.

igalic:
@oysteing only one more thing: could you please [squash](http://www.git-scm.com/book/en/Git-Tools-Rewriting-History#Squashing-Commits) this down to one commit?
****
n.b.: if you use the [github flow](https://guides.github.com/introduction/flow/index.html) work-flow this might be easier or less confusing
see also: https://github.com/voxpupuli/puppet-jira/pull/99#issuecomment-152528108 for a step-by-step how-to.

PR #114, opened 2016-01-22
adding ability to change tomcatShutdownPort

disassembler:
When JIRA is ran on a server with existing tomcat instance, the shutdown port conflicts (hardcoded to 8005). This PR adds a param to init.pp to override the default.


juniorsysadmin:
Looks okay to me. @disassembler  would you mind adding an appropriate RSpec test as well in ```jira_config_spec.rb``` and then doing a force push on your branch?

jyaworski:
This is simple enough that I'm willing to merge without a test just to get this in.

PR #115, opened 2016-01-29
Pe 431 facts

rigareau:
Since upgrading to 2015.3 the jira_fact is in error because $::puppetversion is not reporting that it's running Puppet Enterprise.

I've created a patch that is looking at the agent version 4.3.x and point the ruby interpreter at the default location for 2015.3.x
Also I've updated the condition to not try to install the json packages under Redhat when using the puppet agent 4.3.x that is used in 2015.3.x

### DIFF
--- a/manifests/facts.pp
+++ b/manifests/facts.pp
@@ -29,7 +29,7 @@ class jira::facts (
   if $::puppetversion =~ /Puppet Enterprise/ {
     $ruby_bin = '/opt/puppet/bin/ruby'
     $dir = 'puppetlabs/'
-  } elsif $::values::aio_agent_version =~ /1\.3\.*[0-9]*/ {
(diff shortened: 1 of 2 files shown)


jyaworski:
Hello:

Can you use $::pe_server_version?

https://tickets.puppetlabs.com/browse/MODULES-2279

rigareau:
No I can't use the $pe_server_version on the agent.  But reading the link you provided in your comment made me realise that the decision should be based on the aio_agent_version fact

bastelfreak:
hi @rigareau, thanks for the PR. Are you still interested in this one? If so, please rebase.

rigareau:
Ok thanks

bastelfreak:
I don't quite understand your answer :cry: Do you want to rebase?

Issue #122, opened 2016-03-02
systemd support for Debian 8 breaks Ubuntu


siwilkins:
Commit 47226b0 enables systemd support for Debian >= 8 based on osfamily and operatingsystemmajrelease, but this breaks ubuntu, where osfamily == 'Debian' and operatingsystemmajrelease is based on the year.

PR #123, opened 2016-03-03
Ubuntu initd

siwilkins:
Fix to https://github.com/voxpupuli/puppet-jira/issues/122. Check that the Debian-variant is not Ubuntu before opting to use systemd rather than initd.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -25,18 +25,28 @@ class jira::params {
             for \"${::osfamily}\" - \"${::operatingsystemmajrelease}\"")
       }
     } /Debian/: {
-      if versioncmp($::operatingsystemmajrelease, '8') >= 0 {
-        $json_packages           = 'ruby-json'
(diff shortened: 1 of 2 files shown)


jyaworski:
Can you rebase?

Also, maybe it would be better to put a case statement under osfamily == Debian.

siwilkins:
Is that the kind of thing you had in mind?

jyaworski:
It is. However, I have another idea that may solve this much more generally (and better). Could you use use <code>service_provider</code> as a trigger instead?

```shell
[root@puppetdb2 ~]# facter -p service_provider operatingsystem operatingsystemmajrelease
operatingsystem => CentOS
operatingsystemmajrelease => 6
service_provider => redhat

[root@rundeck ~]# facter -p service_provider operatingsystem operatingsystemmajrelease
operatingsystem => CentOS
operatingsystemmajrelease => 7
service_provider => systemd
```

jyaworski:
This is fine. I'll raise an issue on my point.

PR #127, opened 2016-03-11
Update based on voxpupuli/modulesync_config

dhoppe:
Please merge. I disabled the check ```variable_contains_upcase``` until we fixed this and have a better code coverage.


jyaworski:
Do we have reason to believe that puppet 4's resource types behave that differently?

dhoppe:
Same problem as with the module [confluence](https://github.com/voxpupuli/puppet-confluence/pull/56#issuecomment-195700914).

PR #131, opened 2016-03-22
new custom https redirect port

ericlaflamme:
When using a proxy or port redirection, I don't want Jira to add the port 8443 to the url. So I need Jira to redirect to port 443 even if the tomcat https port is 8443 and then I manage the 443 myself to redirect to 8443. Simple change and defaults is the same as before.

In my case: $tomcat_redirect_https_port       =  443


ericlaflamme:
Sorry, doesn't work

PR #133, opened 2016-03-22
New custom https redirect port

ericlaflamme:
When using a proxy or port redirection, I don't want Jira to add the port 8443 to the url. So I need Jira to redirect to port 443 even if the tomcat https port is 8443 and then I manage the 443 myself to redirect to 8443. Simple change and defaults is the same as before.

In my case: $tomcat_native_ssl = true and $tomcat_redirect_https_port       =  443

We can discuss how to require native_ssl if redirect_https_port is set?

### DIFF
--- a/README.md
+++ b/README.md
@@ -386,6 +386,10 @@ Enable https/ssl support. Defaults to 'false'. See https://confluence.atlassian.
 
 https/ssl Port to listen on, defaults to 8443.
 
+#####`$tomcat_redirect_https_port`
+
+Specifiy Jira redirect https port when using port redirection 80->8080 and 443->8443 or proxy server in front, defaults to $tomcat_https_port. To be used with tomcat_native_ssl.
+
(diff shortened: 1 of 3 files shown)


igalic:
i'm confused about lack of `tomcat_redirect_http_port`, but then again, i suppose we want to encourage secure installations.

****
as far as `native_ssl` goes, i'd add validation:
```puppet
if $tomcat_redirect_https_port {
  validate_integer($tomcat_redirects_port)
  unless ($native_ssl) {
    fail("reasoning…")
  }
}
```

ericlaflamme:
Thanks for the validation code! 👍 
In my case, I configured Jira to redirect in https. So when I first hit 80 -> 8080, is goes to ssl 443. I'm not sure if the URL is changed to jira:8080. I think there is no redirect option in configuration for a normal http port. Maybe I'm wrong but a default config doesn't include a redirect option for normal http.

PR #148, opened 2016-06-02
Added hiera support for catalina_opts - useful for configuring jmx as…

berthamilton:
… java_ops is reused for shutdown and therefore fails due to the port being inuse

### DIFF
--- a/README.md
+++ b/README.md
@@ -315,6 +315,10 @@ defaults to '-XX:-HeapDumpOnOutOfMemoryError'
 
 defaults to ''
 
+#####`$catalina_opts`
+
+defaults to ''
+
 ####Miscellaneous  parameters####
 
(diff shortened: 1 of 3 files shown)


bastelfreak:
Hi @berthamilton , thanks for PR! Can you add a test for this as well?

berthamilton:
Not sure how to do that. I just copied what was there for java_opts and
"tested" that it worked

berthamilton:
I have searched through the existing puppet module and I can't find an example where java_opts or similar has been tested. catalina_opts duplicates this exactly this functionality but it only gets used when tomcat is started - not when stopping the service (which is why you get port errors etc configuring jmx using java_opts). I'm sorry but I'm really the wrong person to add tests here as I don't know ruby.

PR #153, opened 2016-07-12
Enhanced support for new products (JIRA Sofware, Core & Servicedesk).

tgeci:
Atlassian publish their artifacts in multiple variants depending on software type and version (Core, Software, Servicedesk etc.). I added some small improvements to build the correct download path in dependency of the given JIRA Version and JIRA Product type. 
Please merge it after you verified it. The current version of your module isn't applicable for the latest JIRA Version.

Best regards
Taulant

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -44,13 +44,34 @@ class jira::install {
 
   # Examples of product tarballs from Atlassian
   # Core                - atlassian-jira-core-7.0.3.tar.gz
+  # Service Desk        - atlassian-servicedesk-3.1.9.tar.gz
   # Software (pre-7)    - atlassian-jira-6.4.12.tar.gz
-  # Software (7 and up) - atlassian-jira-software-7.0.4-jira-7.0.4.tar.gz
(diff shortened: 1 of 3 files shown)


manifests/install.pp
```diff
@@ -44,13 +44,34 @@
...
+  if $jira::product_type {
+
+    case $jira::product_type {
+
+      'servicedesk':{
+          $file = "atlassian-${jira::product_type}-${jira::version}.${jira::format}"
+      }
+      'core','software':{
+        if ! $jira::latest_version{
+          $file = "atlassian-${jira::product}-${jira::product_type}-${jira::version}-${jira::product}-${jira::version}.${jira::format}"
```
micah-uber: This is not true either. They changed the file naming for 7.1.9 to atlassian-jira-software-7.1.9.tar.gz

igalic: ~_~

tgeci: @micah-uber , @igalic : Sorry for late reaction. Is there a problem with this line ?

igalic: we don't really know

PR #156, opened 2016-08-14
FIX: Download - version 7.1.9 and up

ffquintella:
Just a small fix since atlassian decided to change the name standard in 7.1.9


ffquintella:
Will do some more tests then re-submit

PR #158, opened 2016-08-24
Detect if running in an AIO context and skip install of json_packages

rigareau:
@ffrank I wasn't aware of the voxpupuli/puppet-extlib#63 at the time I wrote this patch.  But it seems to describe a similar issue.


ffrank:
Hi, thanks for the patch!

Is this related? https://github.com/voxpupuli/puppet-extlib/issues/63

Could we move this detection to a common function from extlib in the future?

metadata.json
```diff
@@ -26,7 +26,7 @@
   "requirements": [
     {
       "name": "pe",
-      "version_requirement": ">=3.7.0 <4.0.0"
+      "version_requirement": ">=3.0.0 <=4.3.1"
```
ffrank: So is the Ruby handling the only aspect keeping this module from being Puppet 4 compatible?

Furthermore, does your patch introduce compat w/ 3.x before 3.7?

rigareau: Yes the Ruby handling is the only issue to be Puppet 4 compatible.  I'm using my patch in my environment (Puppet Enterprise 2015.2) for over a month and everything is working smooth.

The way Ruby is installed for version 3.0.0 up to version 3.7 has not changed...

manifests/facts.pp
```diff
@@ -27,10 +27,14 @@
 
   # Puppet Enterprise supplies its own ruby version if your using it.
   # A modern ruby version is required to run the executable fact
+  # Since PE 2015.2 the agents are All in One (AIO) wich also include it's own ruby version
   if $::puppetversion =~ /Puppet Enterprise/ {
     $ruby_bin = '/opt/puppet/bin/ruby'
     $dir      = 'puppetlabs/'
-  } else {
+  } elsif $::aio_agent_version =~ /1\.3\.*[0-9]*/ {
```
petems: Shouldn't this just be `if $::aio_agent_version` rather than just restricting it to 1.3.X?

manifests/facts.pp
```diff
@@ -46,7 +50,8 @@
     }
   }
 
-  if $::osfamily == 'RedHat' and $::puppetversion !~ /Puppet Enterprise/ {
+  # Install $json_packages only if osfamily is RedHat and agent is version PE 3.8 and less or not an AIO version 1.3.x
+  if $::osfamily == 'RedHat' and (($::puppetversion !~ /Puppet Enterprise/) and ($::aio_agent_version !~ /1\.3\.*[0-9]*/)) {
```
petems: Same here, shouldn't this be `if not $::aio_agent_version`?

metadata.json
petems: For PE, now the PE agent and the POSS agent are the same this is generally not needed. I believe we're thinking about deprecating this field entirely as it can cause some confusion for modules.

manifests/facts.pp
rigareau: I hesitated on this one.
But I guess that It's a safe to assume that Ruby will be included in an "All In One" version of a puppet agent...

manifests/facts.pp
rigareau: Here I wasn't sure where the "Puppet Enterprise" comparaison was coming from, so I decided in the name of compatibility to keep that test.

rnelson0:
@rigareau can you rebase this and remove the metadata changes? There's a merge conflict and we can update the metadata before releasing a new version to the forge in a separate PR.

rigareau:
Ok I'll try that

manifests/facts.pp
```diff
@@ -1,61 +1,67 @@
...
+  }
+
+  file { "/etc/${dir}facter/facts.d/jira_facts.rb":
+    ensure  => $ensure,
+    content => template('jira/facts.rb.erb'),
+    mode    => '0500',
+  }
+
+}
+q
```
juniorsysadmin: q

juniorsysadmin:
@rigareau Merge conflicts

rigareau:
The merge conflict has been fixed.

Travis build fails because it can't download the jdk from Oracle...  tried to trigger another build but same issue with jdk download.

ffrank:
I just looked at one run at random and found errors like this one:
```
rspec ./spec/classes/jira_facts_spec.rb[1:1:1:1] # jira::facts supported operating systems on centos-6-x86_64 should contain File[/etc/facter/facts.d/jira_facts.rb]
```
Is this related to the JDK download problem you mentioned?

dpisano:
It looks like this PR has stalled. I have previously submitted a PR to fix this issue, but agreed to have it closed in favor of this PR. I would be more than happy to resubmit it. The only problem would be if I would need to add tests as I don't have the background to be able to add one.

rigareau:
Hi @dpisano please go ahead. I had to find an alternative because it took too much time to resolve this issue.  And since my environment has been moved to the PE 2019 I'm not even sure that the code I've written would still work...

ekohl:
https://github.com/voxpupuli/puppet-jira/commit/6364236b8a9296f527647fd5738254772f6c1e00 fixes this.

Issue #162, opened 2016-09-14
Parameter 'pool_test_on_borrow' does not affect mysql dbconfig.xml


ryandb:
### Affected Puppet, Ruby, OS and module versions/distributions
- Puppet:
- Ruby:
- Distribution:
- Module version:
### How to reproduce (e.g Puppet code you use)

```
class { '::jira':
  version     => $version,
  installdir  => '/usr/local/atlassian/jira',
  homedir     => '/home/jira',
  user        => 'jira',
  uid         => 41,
  group       => 'jira',
  gid         => 41,
  service_manage => false,
  db          => 'mysql',
  dbtype      => 'mysql',
  dbdriver    => 'com.mysql.jdbc.Driver',
  dbuser      => $dbUser,
  dbpassword  => $dbPass,
  dbserver    => 'localhost',
  dbname      => 'jira',
  dbport      => '3306',
  mysql_connector_manage    => true,
  mysql_connector_version   => $sql_connector_version,
  mysql_connector_install   => '/usr/local/atlassian/mysql-java-connector',
  pool_test_on_borrow       => false,
  javahome    => '/opt/java/java',
   proxy          => {
      scheme       => 'https',
      proxyName    => $url,
      proxyPort    => '443',
   },
   ajp       => {
      port            =>         '8009',
      redirectPort    =>        '8443',
      protocol        =>        'AJP/1.3'
  },
} 
```
### What are you seeing

pool_test_on_borrow setting is not in dbconfig.xml
### What behaviour did you expect instead

The following should be in dbconfig.xml:
    <pool-test-on-borrow>false</pool-test-on-borrow>

shollingsworth:
Bump

bastelfreak:
The PR got merged so I think we can close this?

ryandb:
thx

PR #163, opened 2016-09-14
fixes issue #162 - add pool_test_on_borrow to dbconfig.mysql.xml.erb

ryandb:

### DIFF
--- a/README.md
+++ b/README.md
@@ -263,7 +263,7 @@ defaults to true
 
 #####`$pool_test_on_borrow`
 
-defaults to true
+defaults to false
 
 ####MySQL Java Connector parameters####
 
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -84,7 +84,7 @@ class jira (
   $pool_remove_abandoned         = true,
   $pool_remove_abandoned_timeout = 300,
   $pool_test_while_idle          = true,
-  $pool_test_on_borrow           = true,
+  $pool_test_on_borrow           = false,
 
   # JVM Settings
   $javahome     = undef,
--- a/templates/dbconfig.mysql.xml.erb
+++ b/templates/dbconfig.mysql.xml.erb
@@ -20,6 +20,7 @@
     <pool-remove-abandoned><%= scope.lookupvar('jira::pool_remove_abandoned') %></pool-remove-abandoned>
     <pool-remove-abandoned-timeout><%= scope.lookupvar('jira::pool_remove_abandoned_timeout') %></pool-remove-abandoned-timeout>
     <pool-test-while-idle><%= scope.lookupvar('jira::pool_test_while_idle') %></pool-test-while-idle>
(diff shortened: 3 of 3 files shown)


sacres:
Since we actually want this always set to 'false', should we not just set it on the dbconfig rather than create a lookup?

ffrank:
According to the `README`, the `pool_test_on_borrow` value defaults to `true`, though. Am I missing something?

sacres:
There is no value set for `pool_test_on_borrow` for the mysql dbconfig.xml currently and since we always want it set to false, rather than check a variable, I thought we might just set it explicity in the dbconfig.mysql.xml.erb

ffrank:
Sorry if I'm being dense, but I still don't see why it's important or even desirable that the `mysql dbconfig` always receives a `false` value.

It was my initial understanding that that was a bug, and that this PR is supposed to fix it, by using the appropriate parameter value rather than ignoring it. Is that wrong?

sacres:
@ffrank No, the PR isn't _wrong_ per se. I was suggesting & asking the community and the submitter if we should just explicitly set it. 
With regard to why it should be set to false: https://confluence.atlassian.com/jirakb/health-check-database-configuration-792636748.html for example. I'm good with the PR, though think if it's a variable we (I'm game to submit another PR myself updating  `puppet-jira/examples/jira_mysql_install.pp` as well, aside from this).

sacres:
Actually, according to: https://confluence.atlassian.com/adminjiraserver071/tuning-database-connections-802592189.html or https://confluence.atlassian.com/jira064/tuning-database-connections-720411760.html ; Atlassian recommends the setting to be 'false' though _if_ unspecified defaults to 'true'  . Perhaps a better approach would be to change the default here: https://github.com/voxpupuli/puppet-jira/blob/master/manifests/init.pp `$pool_test_on_borrow = true,` to recommended of 'false' .

ryandb:
@sacres - agree wrt defaulting to false.  I will revise the PR.  Thanks.

sacres:
Nice, @ryandb ! We should probably amend the `README` , https://github.com/voxpupuli/puppet-jira/blob/master/README.md#L266 to reflect that change as well though.

ffrank:
Do keep in mind, though, that we will likely need a major release if we want to do this kind of API change.

cc @mkrakowitzer

ryandb:
Changed readme as suggested -- thanks.

sacres:
LGTM @ryandb . What needs to be done to promote the major release you mentioned @ffrank ?

ffrank:
Essentially, if this gets merged to master, we just need to be careful to not create another point release that includes this.

In case there are some things still in the pipeline that also need a major release, it might make sense to hold this back until those are ready. I don't have a good overview over this module. Hope someone will speak up with some authority.

When in doubt, we can always just go ahead and make the version jump. But I'd like a review from at least one other maintainer.

bastelfreak:
looks good to me, discussed it with @sacres on IRC.

PR #177, opened 2016-10-25
Add missing badges

dhoppe:


README.md
```diff
@@ -1,9 +1,13 @@
-#JIRA Module
+# Jira module for Puppet
```
alexjfisher: I think JIRA should probably be all upper-case.

PR #178, opened 2016-10-27
adding access log format variable to init class

pelleqelle:
this one should be more clean and without my git mess :)


pelleqelle:
Hey @bastelfreak anything I can do to move this along?

bastelfreak:
thanks for the PR @pelleqelle !

PR #181, opened 2016-11-04
Major RELEASE PR

sacres:
Since we have made changes to init class as well as a number of fixes that block module usage, I'm proposing a major release with this PR.


bastelfreak:
please add 'modulesync with latest Vox Pupuli defaults'

bastelfreak:
please also add which of the changes is responsible for the major release

Issue #173, opened 2016-10-18
Error: /Stage[main]/Jira::Install/Archive[/tmp/atlassian-jira-software-7.2.3.tar.gz]/ensure: change from absent to present failed: Download file checksum mismatch


felipe1982:
I am getting an error

**Error: /Stage[main]/Jira::Install/Archive[/tmp/atlassian-jira-software-7.2.3.tar.gz]/ensure: change from absent to present failed: Download file checksum mismatch**

When trying to run the below manifest on RHEL 7 Vagrant. `puppet-archive` is version 1.1.2.

``` puppet

class { 'jira':
  javahome => '/opt/java',
  db       => 'postgresql',
  version  => '7.2.3',
}
#  uses tylerwalts-jdk_oracle
include profiles::jdk8
include ::jira::facts
include ::postgresql::server
```

dpisano:
I had to add the checksum to my jira class to get it to work.

dhoppe:
I think this happens since we use archive as default, because we set `checksum` to `undef` but `checksum_verify` defaults to `true`. 

We need to set the checksum for version 6.4.1 and fix the documentation / rspec tests.

felipe1982:
How did you obtain the checksum in the first place?

dpisano:
I downloaded the archive file locally and computed the md5 checksum.

dhoppe:
Unfortunately they do not provide the checksum or a file to verify against. 
https://jira.atlassian.com/browse/JRA-60845

felipe1982:
How do we set `checksum_verify` to `false` ?

sacres:
I was wondering why I hadn't come across this nor could reproduce it. Rather than the archive module, use the deploy module (as mentioned in the [README](https://github.com/voxpupuli/puppet-jira#upgrades-to-the-jira-puppet-module) )
`deploy_module          => 'deploy',`

sacres:
Edited last comment with the _correct_ param. x_X
`deploy_module => 'staging',`

sacres:
@felipe1982 Have you resolved this as recommended?

felipe1982:
@sacres : thanks for checking in. I have not yet tried `staging`, but I do believe that this module does *not* work out of the box to install Jira.  I'll give it a try in a few weeks when I get a chance to play/use/upgrade JIRA again at my office.

Thanks.

sacres:
[...]

PR #182, opened 2016-11-20
init.pp should have 'staging' set as the default as specified

sacres:
The [README](https://github.com/voxpupuli/puppet-jira#upgrades-to-the-jira-puppet-module) specifies that staging is the default deploy module but archive is still specified within init.pp as default.

This PR corrects that.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -101,7 +101,7 @@ class jira (
   $disable_notifications = false,
 
   # Choose whether to use puppet-staging, or puppet-archive
-  $deploy_module = 'archive',
+  $deploy_module = 'staging',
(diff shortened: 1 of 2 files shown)


sacres:
This should also help to avoid issues such as: #173

bastelfreak:
Hi @sacres, thanks for the PR. I'm not sure if it is a good idea to include staging. We are trying to get rid of it. The better solution is probably to fix the archive implementation in this module and not to include staging.

sacres:
Thanks for the info @bastelfreak . I'm following your suggestion re: archive and closing this in favor of new PR #183 .

PR #183, opened 2016-11-20
Atlassian doesn't provide checksums, add checksum_verify false.

sacres:
As well, tests specify jira 6.0.0, it's 6.0

Fix jira 6.0 source/download urls

fix jira 6 version as well...

Changing 6.0 to 6.1 as less then 6.1 is EOL

match jira 6.1 source and download urls

While fixing tests for Jira 6.0, I noticed that versions < 6.1 are end of support by Atlassian:
[End Of Support](https://confluence.atlassian.com/jira064/supported-platforms-720411857.html)
I changed the test to use 6.1 instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,9 +101,8 @@ The jira::facts class is required for upgrades.
 
 ##### Upgrades to the JIRA puppet Module
 
-mkrakowitzer-deploy has been replaced with nanliu-staging as the default module for
-deploying the JIRA binaries. You can still use mkrakowitzer-deploy with the
-*deploy_module => 'archive'*
+puppet-archive is the default module for
+deploying the JIRA binaries.
 
 ```puppet
   class { 'jira':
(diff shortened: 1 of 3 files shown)


sacres:
This should fix the issue described in #173

sacres:
@bastelfreak This PR follows your suggestion that we are moving towards deprecating puppet-staging and using puppet-archive . Thanks again for the review and info!

spacepants:
Dunno about other folks, but I really prefer having checksum verification as it lets me ensure that I have a complete and working archive. Can we not just update the default checksum and version if that's an issue?

sacres:
@spacepants Well,  the vendor has to supply the checksum and they do not.

bastelfreak:
Thanks for the fix @sacres! @spacepants I would love to turn on checksum verification, but it isn't possible because atlassian doesn't provide anything. It is currently activated in this module, yes, but thats a bug.

PR #187, opened 2016-11-26
Fix Beaker setup because of the migration to puppet-archive

dhoppe:


spec/acceptance/default_parameters_spec.rb
```diff
@@ -32,15 +32,22 @@ class { 'postgresql::globals':
...
+        extract_command => 'tar xfz %s --strip-components=1',
+        extract_path    => $jh,
+        source          => "#{java_url}",
+        creates         => "${jh}/bin",
+        cleanup         => true,
+        cookie          => 'oraclelicense=accept-securebackup-cookie',
       } ->
       class { 'jira':
         version      => '6.2.7',
+        checksum     => 'ddacb69b9327bd40ab6ce726f0dd4a08',
```
sacres: @dhoppe Thanks, nice work. I successfully tested via beaker 
As we discussed on IRC regarding the current lack of of hashes [md5, sha] provided by Atlassian; can we remove the checksum lines?
LGTM other than those two lines

dhoppe: I would like to change the `download_url` from `http` to `https`, even if Atlassian redirects the traffic and provide the `checksum` for our `version`. This way the module should work out of the box. 

But I do not have enough time to update the Atlassian modules to the latest version. This would require more testing regarding the facts and that Confluence 5.8 dropped the support of Java 7 etc.

sacres: Sounds reasonable. Perhaps I could update them for this module to match what we have in the spec/class tests. If that's the case, let's merge this (though I'm still thinking that the checksum lines should be removed).

juniorsysadmin:
I know the fix was specifically for puppet-archive, but we should also make sure we test against JDK8 not 7 in another PR

spec/acceptance/default_parameters_spec.rb
```diff
@@ -19,7 +19,7 @@
   it 'installs with defaults' do
     pp = <<-EOS
       $jh = $osfamily ? {
-        default   => '/opt/java',
+        default => '/opt/java',
       }
       if versioncmp($::puppetversion,'3.6.1') >= 0 {
```
juniorsysadmin: We no longer need this given metadata has 3.8.7 as minimum Puppet version

juniorsysadmin:
Superseded by #197

PR #188, opened 2016-11-30
Add checksum for modules default version

dhoppe:
As long as Atlassian does not provide checksums, we should at least supply one for the default version so that the installation is as straightforward as possible.


dhoppe:
Closed in favour of https://github.com/voxpupuli/puppet-jira/pull/188.

PR #189, opened 2016-11-30
Use https instead of http

dhoppe:
We should use https instead of http, even if Atlassian redirects the traffic automatically.

### DIFF
--- a/README.md
+++ b/README.md
@@ -370,7 +370,7 @@ defaults to ''
 ##### `$download_url`
 
 The URL used to download the JIRA installation file.
-Defaults to `https://downloads.atlassian.com/software/jira/downloads/`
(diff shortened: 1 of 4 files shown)


bramwelt:
This would actually revert my PR #110.

+1 to fixing the URL in the README and tests to match the default.
-1 to reverting back to 'www.atlassian.com'.

bramwelt:
Also the `jira::download_url` default does use HTTPS.

sacres:
@dhoppe The URLs _are_ `'https' `now. Can we close this please?

dhoppe:
At the moment https is used, but what if Atlassian changes it again or the web server is configured incorrectly? Why do we actually have three modules from Atlassian which are all configured differently? Nobody has complained about the other modules use https.

PR #191, opened 2016-12-03
Exposes checksum_verify in init.pp w/ conditional.

sacres:
This PR addresses the concerns of my previous (incorrect) hardcoding of checksum_verify (as the default for that is 'true' in archive module) in install.pp and exposes it with a conditional in init.pp instead. It sets the value depending on whether checksum is defined by user and performs the check, otherwise it does no check. 
@dhoppe @afisher @bbriggs @bastelfreak (and anyone else), please review. Thanks!

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -87,13 +87,13 @@ class jira (
   $pool_test_on_borrow           = false,
 
   # JVM Settings
-  $javahome     = undef,
-  $jvm_xms      = '256m',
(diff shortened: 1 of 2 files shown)


dhoppe:
@sacres I would love to see this kind of change for Confluence and Stash as well.

sacres:
I'm willing @dhoppe 
Thanks all!

PR #192, opened 2016-12-06
Allow management of vendor check-java.sh file

bramwelt:
The script which checks for Java assumes Oracle Java, and thus wrongly
checks the version string output to be prefixed with 'java' and not
'openjdk'. This change removes the check for JIRA 7 and up.

Signed-off-by: Trevor Bramwell <tbramwell@linuxfoundation.org>

### DIFF
--- a/README.md
+++ b/README.md
@@ -418,10 +418,8 @@ Defaults to an empty string (""). Will add a path to the Tomcat Server Context.
 
 ##### `$script_check_java_manage`
 
-Manages a modified version of the 'check-java.sh' script provided by
-JIRA in order to use OpenJDK.
-
-Note: JIRA does not officially support OpenJDK.
+Manages a the 'check-java.sh' script provided by
+JIRA.
 
 Defaults to 'false'.
 
--- a/README.md
+++ b/README.md
@@ -416,6 +416,22 @@ Defaults to {}, See examples on how to use.
 
 Defaults to an empty string (""). Will add a path to the Tomcat Server Context.
 
+##### `$script_check_java_manage`
+
+Manages a modified version of the 'check-java.sh' script provided by
+JIRA in order to use OpenJDK. (Requires JIRA > 7.0.0)
+
+Note: JIRA does not officially support OpenJDK.
+
+Defaults to 'false'.
+
+##### `$script_check_java_template`
+
+Alternate location to find the 'check-java.sh' script. Requires
+`$manage_script_check_java = true`.
+
+Defaults to undef.
+
 #### Tomcat parameters
 
 ##### `$tomcat_address`
--- a/README.md
+++ b/README.md
@@ -419,7 +419,7 @@ Defaults to an empty string (""). Will add a path to the Tomcat Server Context.
 ##### `$script_check_java_manage`
 
 Manages a modified version of the 'check-java.sh' script provided by
-JIRA in order to use OpenJDK. (Requires JIRA > 7.0.0)
+JIRA in order to use OpenJDK.
 
 Note: JIRA does not officially support OpenJDK.
 
@@ -428,9 +428,9 @@ Defaults to 'false'.
 ##### `$script_check_java_template`
 
 Alternate location to find the 'check-java.sh' script. Requires
-`$manage_script_check_java = true`.
+`$script_check_java_manage = true`.
 
-Defaults to undef.
+Defaults to 'jira/check-java.sh.erb'.
 
 #### Tomcat parameters
 
--- a/files/check-java.sh
+++ b/files/check-java.sh
@@ -1,0 +1,23 @@
+
+#!/bin/sh
+
+_EXPECTED_JAVA_VERSION="8"
+
+#
+# check for correct java version by parsing out put of java -version
+# we expect first line to be in format 'version "1.8.0_40"' and assert that minor version number will be 8 or higher
+#
+
+"$_RUNJAVA" -version 2>&1 | grep "version" | (
+        IFS=. read ignore1 version ignore2
+        if [ ! ${version:-0} -ge "$_EXPECTED_JAVA_VERSION" ]
+        then
+           echo "*************************************************************************************************************************************"
+           echo "**********     Wrong JVM version! You are running with "$ignore1"."$version"."$ignore2" but JIRA requires at least 1.8 to run.      **********"
+           echo "*************************************************************************************************************************************"
+           exit 1
+        fi
+    )
+if [ $? -ne 0 ] ; then
+   exit 1
+fi
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -54,15 +54,22 @@ class jira::config inherits jira {
     notify  => Class['jira::service'],
   }
 
-  if (versioncmp($jira::version, '7.0.0') >= 0) {
-    file { "${jira::webappdir}/bin/check-java.sh":
-      source  => "puppet:///modules/${module_name}/check-java.sh",
-      mode    => '0755',
-      require => [
-        Class['jira::install'],
-        File["${jira::webappdir}/bin/setenv.sh"],
-      ],
-      notify  => Class['jira::service'],
+  if $jira::script_check_java_manage {
+    if $jira::script_check_java_template == undef {
+      $script_check_java_location = "puppet:///modules/${module_name}/check-java.sh"
+    } else {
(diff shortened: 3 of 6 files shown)


sacres:
@bramwelt Thanks for your work on this PR.
I'll note that Atlassian doesn't support openjdk.
Also, although this module _needs_ java; it doesn't actually have a dependancy upon another module covering that need inately.
I do agree that there are versions of the officially vendor supported java which are needed (i.e. jira < 6.3 won't run on jdk 8) etc. but openjdk isn't one of them (working or not; it's unsupported upstream).
Thoughts @dhoppe and @bastelfreak ?
Thanks!

juniorsysadmin:
As mentioned on IRC I think the issue is less about trying to 'support' OpenJDK, and more about allowing flexibility.

A nicer course of action for this module would be to:
1) Allow the user to choose their own template for those scripts individually, or with a boolean choose not to manage certain files individually
2) Keep the scripts that have been made into a template reasonably up-to-date with what is in the tarball of newer versions (I doubt there’s ever been an attempted ‘sync’)

juniorsysadmin:
So, in keeping with the above points I would prefer a parameter boolean ```manage_script_check_java = false``` which changes whether we manage the file resource, plus a ```script_check_java_template = undef``` parameter which allows one to provide a template path if required.

bramwelt:
@sacres It's true that Atlassian does not support OpenJDK, and I'm aware of that. I'm not sure how module dependencies, and specific Java version (7 vs. 8) is related to this change? Surprisingly this script is the one thing needed to make JIRA run on OpenJDK.

@juniorsysadmin wrote:
> Allow the user to choose their own template for those scripts individually, or with a boolean choose not to manage certain files individually

I completely agree. I did not feel completely comfortable laying down a file that is technically put in place by the application 'as code', and I think making this optional is a good idea.

> script_check_java_template = undef parameter which allows one to provide a template path if required.

That seems far too reasonable and flexible of a request. ;)

Thanks for the guidance. I'll update the PR.

sacres:
@bramwelt My comments re: 7/8 java versions are more toward the lack of any checking for java versions (within the module) and I shouldn't have bothered mentioning that in this PR.

bramwelt:
@sacres No worries, and thanks for clarifying.

opnfv-github:
@juniorsysadmin In order to match the style of the module I used `$script_check_java_manage` and `$script_check_java_template`, but the logic you requested is still the same.

bramwelt:
@juniorsysadmin In order to match the style of the module I used `$script_check_java_manage`, but the logic is still the same.

manifests/config.pp
```diff
@@ -52,12 +52,35 @@
     mode    => '0755',
     require => Class['jira::install'],
     notify  => Class['jira::service'],
-  } ->
+  }
+
+  if $jira::script_check_java_manage {
```
juniorsysadmin: I was thinking of something much simpler:
```puppet
if $script_check_java_manage {
  file { "${jira::webappdir}/bin/check-java.sh":
    content => template($jira::script_check_java_template),
    # Or just $jira::script_check_java_template if you require users to put 'file(path-to-file)'
    # or 'template(path-to-file) for the $script_check_java_template parameter
    mode    => '0755',
    require  => [
      Class['jira::install'].
      File["${jira::webappdir}/bin/setenv.sh"],
    ],
   notify  => Class['jira::service'],
  }
}
```

manifests/init.pp
```diff
@@ -113,6 +113,10 @@
   # incase the jira service is managed outside of puppet. eg: using the
   # puppetlabs-corosync module: 'crm resource stop jira && sleep 15'
   $stop_jira = 'service jira stop && sleep 15',
+  # Whether to manage the 'check-java.sh' script, and where to retrieve
+  # the script from.
+  $script_check_java_manage = false,
+  $script_check_java_template = undef,
```
juniorsysadmin: If we're checking in a duplicate of the file in the tarball (see above) we can make this:
```puppet
$script_check_java_template = 'template("${module_name}/check-java.sh.erb")'
# or
$script_check_java_template = "${module_name}/check-java.sh.erb"
```

manifests/config.pp
```diff
@@ -52,12 +52,35 @@
...
+        notify  => Class['jira::service'],
+      }
+    }
+  }
 
   file { "${jira::homedir}/dbconfig.xml":
     content => template("jira/dbconfig.${jira::db}.xml.erb"),
     mode    => '0600',
-    require => [ Class['jira::install'],File[$jira::homedir] ],
+    require => [
```
bramwelt: This was needed because of my removal of `->` after the `setenv.sh` script management. I don't recall why I decided to put the code block here, so I'll revert this back.

README.md
```diff
@@ -416,6 +416,22 @@ Defaults to {}, See examples on how to use.
...
+JIRA in order to use OpenJDK. (Requires JIRA > 7.0.0)
+
+Note: JIRA does not officially support OpenJDK.
+
+Defaults to 'false'.
+
+##### `$script_check_java_template`
+
+Alternate location to find the 'check-java.sh' script. Requires
+`$manage_script_check_java = true`.
```
bramwelt: Done.

files/check-java.sh
```diff
@@ -0,0 +1,23 @@
+
```
bramwelt: Yes this does contain my version. I've updated the file with comments and added the 'managed by Puppet' header.

manifests/config.pp
bramwelt: Done.

manifests/init.pp
bramwelt: I don't believe using `'template(...)'` as a string will work, as _template_ is a function, and the `content` attributes of files doesn't parse the string. So this either has be set to `file` or `template` in the content attribute of the file resource, and I opted for `template` as that is what you originally requested.

Given the file doesn't use any variables, it doesn't make much sense, but at least it seems consistent with the rest of the module?

templates/check-java.sh.erb
```diff
@@ -0,0 +1,28 @@
+#!/bin/sh
+# This file is managed by Puppet
+#
+# This file should be identical to 'check-java.sh' in the Jira 7.0.4 standalone
+# tarball, with 2 exceptions: 'grep "java version"' has been changed
+# to 'grep "version"' in order to allow JIRA to run on OpenJDK, and the
```
juniorsysadmin: You might think of it as a bit a nitpick, but referring back to the earlier discussion I still think the grep command should be left exactly like the one in the tarball. That way:
- You stick a modified version of the template with the modified grep command in say my_sites/templates/jira/check-java.sh.erb in your module directory (1)
- You pass in an argument as a parameter to this module of ```script_check_java_template => 'my_sites/jira/check-java.sh.erb'```
- Everyone else using this module by default gets the same file as the one that's in the upstream tarball
- Anyone looking at this template sees that we're just taking what is upstream and not doing something which seems weird
- Noone comes along and questions why there is a reference to OpenJDK in the code

Everything else in the PR looks okay.

(1) The my_sites directory in your module directory used to be the 'standard' folder name for the folder containing all of your modified templates. Not sure if there is a more catchy name for it now

bramwelt: Understood.

I've updated the change to use a vanilla `check-java.sh` script. :slightly_smiling_face:

README.md
```diff
@@ -416,6 +416,20 @@ Defaults to {}, See examples on how to use.
 
 Defaults to an empty string (""). Will add a path to the Tomcat Server Context.
 
+##### `$script_check_java_manage`
+
+Manages a the 'check-java.sh' script provided by
```
juniorsysadmin: Typo

igalic:
i'm very confused as to why this check is necessary to begin with?
we depend on @puppetlabs java, which contains various facts for the java version.

juniorsysadmin:
> i'm very confused as to why this check is necessary to begin with?
we depend on @puppetlabs java, which contains various facts for the java version.

The Jira tarball comes with a bunch of scripts that are used for startup and shutdown of the service out of the box. This PR allows management of one of those files by Puppet. For this particular case we probably don't need to check a template into the code since no Ruby evaluation is being performed in the template - ```script_check_java_template = undef``` would have been fine. But I'm fine leaving it in as a model of how we should handle the other scripts that this module overwrites

PR #198, opened 2017-01-03
FIX version checking regex 

sacres:
Fairly sure I had amended this before but at any rate that didn't make it into the last PR #197. Most likely (PEBCAK) 😈 
This fixes:
- The version check doesn't match version


sacres:
Thanks for the review @juniorsysadmin ! Small but vital change which is clear; merging.

PR #205, opened 2017-01-13
Jira7.1 redirect to https

ericlaflamme:
I opened this pull request to start the discussion.

At the moment, I set redirect https only if:

you set the version variable to 7.1.0 and up
and
if you set both tomcat_redirect_https_port and tomcat_native_ssl

It's safe to say that 7.1.0 accept this config and I dont' want to mess the web.xml file if you don't want to redirect. But I would like to know how we determine we need to redirect. I'm pretty sure that you might want to redirect even without $tomcat_redirect_https_port?

Is only tomcat_native_ssl is enough?

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -72,7 +72,14 @@ class jira::config inherits jira {
       notify  => Class['jira::service'],
     }
   }
-
+  if $jira::tomcat_redirect_https_port and $jira::tomcat_native_ssl {
+    file { "${jira::webappdir}/atlassian-jira/WEB-INF/web.xml":
+      content => template('web.xml.erb'),
+      mode    => '0644',
+      require => Class['jira::install'],
+      notify  => Class['jira::service'],
+    }->
+  }
   file { "${jira::webappdir}/conf/server.xml":
     content => template('jira/server.xml.erb'),
     mode    => '0600',
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -72,14 +72,18 @@ class jira::config inherits jira {
       notify  => Class['jira::service'],
     }
   }
-  if $jira::tomcat_redirect_https_port and $jira::tomcat_native_ssl {
-    file { "${jira::webappdir}/atlassian-jira/WEB-INF/web.xml":
-      content => template('web.xml.erb'),
-      mode    => '0644',
-      require => Class['jira::install'],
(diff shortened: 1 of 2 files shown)


sacres:
Hey @ericlaflamme . I don't see why you're referencing web.xml, it is not the config you are looking for.

ericlaflamme:
Hi @sacres , I'm not sure I follow your comment. I want to make redirect to https work on Jira 7.1 and web.xml is the config file you need to modified to make it work. It came from the official documentation of atlassian:
https://confluence.atlassian.com/adminjiraserver071/running-jira-applications-over-ssl-or-https-802593051.html#RunningJIRAapplicationsoverSSLorHTTPS-Advancedconfiguration

sacres:
Hi @ericlaflamme . Hmm, yes indeed, all 8277 lines of it 😲 
I think we should manage this config file differently. From `WEB-INF/web.xml:`
```
<!--
All session-config, mime-mapping, welcome-file-list, error-page, taglib,
resource-ref, security-constraint, login-config, security-role,
env-entry, and ejb-ref elements should follow this fragment.
-->

**FOO BLOCK HERE**

</web-app>
```
Would you be able to work out something which inserts the required block @ **FOO BLOCK HERE**  and if so,  also amend the `README.md` to reflect this.
Thanks!

ericlaflamme:
Hi @sacres, I agree with you that 8000+ lines is obviously not the best thing but what do you suggest? The thing is if we use something like Augeas, we will end up with an additionnal dependancy for injecting 8 lines.

Puppet isn't really good at injecting text.. Puppet is base by the fact that we manage the whole file or not at all.. And then some modules want to solved that but isn't that obvious.

Also, I'm not sure what to modified in the readme since it fixed an already working feature that isn't working anymore on jira 7.1. So it basically make the module work again :)

oranenj:
I really don't want another template in the module that needs to be kept up-to-date with upstream.

This is an understandable feature request, but the price seems way too steep when you could just put a proxy in front of JIRA :/

oranenj:
Given that this PR is over 3 years old and I'm just about to drop support for pre-8.0 JIRA from the module, I don't think there's value in keeping it open.

If people want to manage server.xml, they are able to do it outside this module. We should focus on making common production setups easy rather than trying to support everything.

PR #210, opened 2017-03-27
Fix tls connector protocol

camino404:
for jira version 7.3, the tls protocol has been changed to : org.apache.coyote.http11.Http11NioProtocol.
This is a quick fix. Do you guys think we should be sending this as a hiera parameter or have an if else statement on the jira version number?

### DIFF
--- a/templates/server.xml.erb
+++ b/templates/server.xml.erb
@@ -73,7 +73,7 @@
 <% if @tomcat_native_ssl -%>
         <Connector
                     port="<%= @tomcat_https_port %>"
-                    protocol="org.apache.coyote.http11.Http11Protocol"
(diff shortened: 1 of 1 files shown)


bastelfreak:
Hi @camino404, please check the used email address in your commit. It isn't associated with your github account.

wouterhummelink:
This is already a parameter for non-tls connector, it should use that (Default value: HTTP/1.1 autoselects implementation)

djmentos:
Maybe moving it to variable will be a better idea?

bastelfreak:
Hi. Thanks for raising this PR, but I'm going to close it due to inactivity. Please reopen it if you're still interested in getting this merged.

PR #212, opened 2017-04-03
replace validate_* with datatypes

bastelfreak:

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -33,156 +33,129 @@
 class jira (
 
   # Jira Settings
-  $version      = '6.4.1',
-  $product      = 'jira',
-  $format       = 'tar.gz',
-  $installdir   = '/opt/jira',
-  $homedir      = '/home/jira',
-  $user         = 'jira',
-  $group        = 'jira',
-  $uid          = undef,
-  $gid          = undef,
-  $shell        = '/bin/true',
-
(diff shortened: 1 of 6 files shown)


bastelfreak:
the goal of this is to get rid of the validate_* calls because they are deprecated. The goal is not to add datatypes to all parameters.

manifests/init.pp
```diff
@@ -33,156 +33,129 @@
...
+  String $version                                                   = '6.4.1',
+  String $product                                                   = 'jira',
+  String $format                                                    = 'tar.gz',
+  Stdlib::Absolutepath $installdir                                  = '/opt/jira',
+  Stdlib::Absolutepath $homedir                                     = '/home/jira',
+  String $user                                                      = 'jira',
+  String $group                                                     = 'jira',
+  $uid                                                              = undef,
+  $gid                                                              = undef,
+  Stdlib::Absolutepath $shell                                       = '/bin/true',
```
alexjfisher: If the aim is to improve readability, I'm not sure aligning *everything* is helpful.

alexjfisher: Whilst googling, I came across this style too. https://github.com/jlambert121/jlambert121-puppet/blob/master/manifests/init.pp

bastelfreak: @alexjfisher maybe we should merge this as is and discuss it on the mailing list?

PR #220, opened 2017-09-30
modulesync 1.2.0

wyardley:
The modulesync had failed so many times; I just deleted the old branch / PR and created this one.
We already put in a fix for the templating thing that was causing the rubocop failure with 1.1.0.

### DIFF
--- a/.msync.yml
+++ b/.msync.yml
@@ -1,1 +1,1 @@
-modulesync_config_version: '0.21.0'
+modulesync_config_version: '1.2.1'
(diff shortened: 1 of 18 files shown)


wyardley:
So, tests are passing now. I did some structural work as well (that maybe should have been outside modulesync), but I think this has stuff at least in a better state before.
Ubuntu still has some slight wackiness in the acceptance tests (though basically works)

.msync.yml
```diff
@@ -1 +1 @@
-modulesync_config_version: '0.21.0'
+modulesync_config_version: '1.2.1'
```
bastelfreak: in the future please don't sync from HEAD, checkout the latest tag and use that instead

PR #221, opened 2017-10-01
bump archive dep to allow latest version

bastelfreak:


ekohl:
Looks like this breaks puppet module install because of deps.

bastelfreak:
We need to wait for puppetlabs-java: https://github.com/puppetlabs/puppetlabs-java/pull/261

PR #229, opened 2017-11-21
Use systemd on Ubuntu 16.04

KoenDierckx:
Update params to use systemd on ubuntu >= 15.04. 
Also updated line ends from crlf to lf for templates, as these gave issues with shell scripts


manifests/params.pp
```diff
@@ -27,11 +27,19 @@
     } /Debian/: {
       case $::operatingsystem {
         'Ubuntu': {
-          $json_packages           = [ 'rubygem-json', 'ruby-json' ]
-          $service_file_location   = '/etc/init.d/jira'
-          $service_file_template   = 'jira/jira.initscript.erb'
-          $service_lockfile        = '/var/lock/jira'
-          $service_provider        = 'debian'
+          if versioncmp($::operatingsystemmajrelease, '15.04') >= 0 {
```
bastelfreak: Can you migrate this to the new facts hash?

bastelfreak:
Hi @KoenDierckx, thanks for the PR. Can you add ubuntu16.04 to the metadata.json file?

bastelfreak:
Thanks for all the work!

PR #252, opened 2018-06-13
Update server.xml.erb to account for weird Service Desk versioning

ronech:
If you set `$product = 'servicedesk'` (which allows you to install the Jira Service Desk application, a superset of Jira Server), its versioning is currently in the 3.x range, which hits this line in `server.xml.erb` and causes the older JasperListener to be called at startup and it will fail.  This has been tested with `$product = 'servicedesk'` and `$version = '3.12.2'`.


bastelfreak:
thanks for the PR @ronech !

PR #256, opened 2018-07-03
Add Support for H2 Database and Fix #73 

### LINKED ISSUE #73: consolidate db, dbtype, and dbdriver
Right now you have to set three configs in order to accomplish the switching between one database type and another. I think it makes far more sense to change the dbtype and dbdriver values based off of what people set for db.

### LINKED ISSUE #254: Add H2 database support

- Puppet: 5.5.2
- Ruby: ruby 2.4.3p205 (2017-12-14 revision 61247) [x86_64-linux]
- Distribution: CentOS 7.5
- Module version: 3.4.1

```
include jira
include jira::facts
```

```
jira::version: 7.10.2
jira::javahome: /usr/java/latest
jira::jvm_xms: 384m
jira::db: h2
```

## What are you seeing
It doesn't accept h2.

```
  Enum['postgresql','mysql','sqlserver','oracle'] $db               = 'postgresql',
```

I expected the "default" (built-in) database to work, for vagrant testing.

Here is the dbconfig.xml from a test deployment with H2:
(obviously the "url"  appears to have a filename /apps/jira-shared/database/h2db )

```
<?xml version="1.0" encoding="UTF-8"?>

<jira-database-config>
  <name>defaultDS</name>
  <delegator-name>default</delegator-name>
  <database-type>h2</database-type>
  <schema-name>PUBLIC</schema-name>
  <jdbc-datasource>
    <url>jdbc:h2:file:/apps/jira-shared/database/h2db</url>
    <driver-class>org.h2.Driver</driver-class>
    <username>sa</username>
    <password></password>
    <pool-min-size>20</pool-min-size>
    <pool-max-size>20</pool-max-size>
    <pool-max-wait>30000</pool-max-wait>
    <min-evictable-idle-time-millis>4000</min-evictable-idle-time-millis>
    <time-between-eviction-runs-millis>5000</time-between-eviction-runs-millis>
    <pool-max-idle>20</pool-max-idle>
    <pool-remove-abandoned>true</pool-remove-abandoned>
    <pool-remove-abandoned-timeout>300</pool-remove-abandoned-timeout>
  </jdbc-datasource>
</jira-database-config>
```

TJM:

Add Support for H2 database and add sensible defaults for other db related params.

Fixes #254 
Fixes #73

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -207,7 +207,7 @@ class jira (
       'postgresql' => 'org.postgresql.Driver',
       'mysql'      => 'com.mysql.jdbc.Driver',
       'oracle'     => 'oracle.jdbc.OracleDriver',
-      'sqlserver'  => 'net.sourceforge.jtds.jdbc.Driver',
+      'sqlserver'  => 'com.microsoft.sqlserver.jdbc.SQLServerDriver',
       'h2'         => 'org.h2.Driver',
     }
   }
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -50,15 +50,15 @@ class jira (
   Boolean $datacenter                                               = false,
   $shared_homedir                                                   = undef,
   # Database Settings
-  Enum['postgresql','mysql','sqlserver','oracle'] $db               = 'postgresql',
+  Enum['postgresql','mysql','sqlserver','oracle','h2'] $db          = 'postgresql',
   $dbuser                                                           = 'jiraadm',
   $dbpassword                                                       = 'mypassword',
   $dbserver                                                         = 'localhost',
   $dbname                                                           = 'jira',
-  $dbport                                                           = '5432',
-  $dbdriver                                                         = 'org.postgresql.Driver',
-  $dbtype                                                           = 'postgres72',
-  $dburl                                                            = undef,
+  Optional[Variant[Integer,String]] $dbport                         = undef,
+  Optional[String] $dbdriver                                        = undef,
+  Optional[String] $dbtype                                          = undef,
+  Optional[String] $dburl                                           = undef,
   $poolsize                                                         = '20',
   $dbschema                                                         = 'public',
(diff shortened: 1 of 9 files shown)


manifests/init.pp
```diff
@@ -50,15 +50,15 @@
...
+  Enum['postgresql','mysql','sqlserver','oracle','h2'] $db          = 'postgresql',
   $dbuser                                                           = 'jiraadm',
   $dbpassword                                                       = 'mypassword',
   $dbserver                                                         = 'localhost',
   $dbname                                                           = 'jira',
-  $dbport                                                           = '5432',
-  $dbdriver                                                         = 'org.postgresql.Driver',
-  $dbtype                                                           = 'postgres72',
-  $dburl                                                            = undef,
+  Optional[Variant[Integer,String]] $dbport                         = undef,
```
bastelfreak: datatypes \o/

templates/dbconfig.h2.xml.erb
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
bastelfreak: could you convert this to an epp() template please?

bastelfreak:
Hi @TJM, thanks for the PR! Can you please take a look at the inline comment I made?

templates/dbconfig.h2.xml.erb
TJM: I figured consistency is better, otherwise we have to convert all the templates to EPP, and if we bother doing that, we should just make it one template and conditionally include the lines as needed instead of having several files that are almost exactly the same.

TJM: @bastelfreak What do you think?

bastelfreak: ah good point, let's keep it as is.

PR #258, opened 2018-07-11
Check `ps` for a process with the jira pid, not a the jira user

### LINKED ISSUE #243: RedHat 6 Service Manifest problem

- Puppet: 2017.3.4
- Ruby: Any
- Distribution: RedHat 6.9
- Module version:3.3.0 - 3.4.1

Install eazybi and configure tu run an external process to execute the reports.  

Or for development and testing purposes you can just look at a log file while being logged in with the userid that is executing jira.
sudo -iu jira
tail -f /home/jira/logs/atlassian-jira.log

Either of theses actions will create a process that is matched by the pgrep expression that is used to detect if the service is running and will throw an error because there is more than 1 process id matching.

## What are you seeing
When I ask the Jira service for a status it produce the following error.
$>service jira status
/etc/init.d/jira: line 66: [: too many arguments
jira is stopped

service jira status
jira is running

## Output log
$>service jira status
/etc/init.d/jira: line 66: [: too many arguments
jira is stopped

Replace service status test to use the catalina.pid content just as in the redhat 7 service unit file.

function status() {

STATUS=$( pgrep -F <%= scope.lookupvar('jira::webappdir') %>/work/catalina.pid ) || STATUS=0
 if [ $STATUS -ne 0 ];then
   echo "$SERVICE is running"
   return 0
 else
   echo "$SERVICE is stopped"
   return 1
 fi

}

rigareau:
RedHat 6 Service Manifest should rely on jira pid file instead of querying if any process is currently running as a specific userid

Fixes #243


bastelfreak:
thanks for fixing this @rigareau !

PR #172, opened 2016-10-14
Added parameter for alternative data_dir

tgeci:
We install our atlassian instances out of the users home dir. So I modified the code to be able to use /opt/atlassian/jira/data as path for the atlassian data. I added a new parameter called jira::data_dir, which is used in config.pp for creating config files. 

Please merge if its ok. I think its also a benefit for other users.

Best regards
Taulant

PR #266, opened 2018-09-26
Allow homedir to be managed from outside of module

timdeluxe:
We run Jira in a cluster setup (pacemaker). For that we need to have the homedir of Jira (which is also the "data" directory) on a NFS share. The NFS mount is done to another directory and the data dir is a symlink, which is managed outside of this jira module. To be able to do that, i needed to fix the jira module to not handle the homedir, if done from outside. I think it will be useful for others, who are building similar setups, hence this pull request.

n/a

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -126,7 +126,7 @@ class jira::install {
     }
   }
 
-  file { $jira::homedir:
+  if ! defined(File[$jira::homedir]) {
     ensure => 'directory',
     owner  => $jira::user,
     group  => $jira::group,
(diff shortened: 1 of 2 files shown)


timdeluxe:
I just saw that #172 is similar to this and still open. Also #172 is pretty much outdated and needs to get rebased against the current codebase. I think these two do not fight against each other. If someone wants to have the data dir separated, one can do so once that PR is merged and if one wants to have the homedir managed separatly, one can do with mine.

bastelfreak:
Hi @timdeluxe, can you take a look at the failing travis jobs?

ekohl:
Is this still something you're interested in or can it be closed?

timdeluxe:
@ekohl We changed away from a NFS to a shared disc / direct mount solution for the cluster, so the already existing code is now okay for us. No need for this PR anymore, closing it now. Thanks for reminding.

PR #272, opened 2018-11-01
Remove messy warnings. Fixes #234

### LINKED ISSUE #234: Cannot add wished tag to server.xml
According to this problem, https://confluence.atlassian.com/jirakb/jira-throws-java-io-filenotfoundexception-warnings-after-upgrade-869007981.html , one has to add a tag inside the server.xml file in order to get rid of the warning messages. By looking at the server.xml.erb template I found out that it is currently not possible to add tags at the wished position (Inside the Context-tag). Therefore I want to ask if the necessary functionality could be added, so this tag could be inserted if required.

br0ch0n:
Seems that per https://confluence.atlassian.com/jirakb/jira-throws-java-io-filenotfoundexception-warnings-after-upgrade-869007981.html the scanner was always off by default, so it should be safe to just set it to false.

Fixes #234

### DIFF
--- a/templates/context.xml.erb
+++ b/templates/context.xml.erb
@@ -41,4 +41,5 @@
     />
 <%   end -%>
 <% end -%>
+<JarScanner scanManifest="false"/>
 </Context>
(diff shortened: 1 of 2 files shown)


bastelfreak:
Thanks for the fix @br0ch0n !

PR #282, opened 2019-02-09
Add Ehcache options; require stdlib 4.25.0

### LINKED ISSUE #245: Support for additional cluster.properties config
It'd be nice to be able to specify additional cluster config settings for the `cluster.properties` file. Such as:
* ehcache.listener.hostName
* ehcache.listener.port
* ehcache.object.port

Right now, manually adding them causes them to be overwritten on next puppet run by the template. 

Current work around is to perform initial provisioning of a Data Center node with `datacenter => true`, then set it to `false` and manually add additional config settings.

nrhtr:
Add additional options for configuring the "ehcache" settings in cluster.properties (https://developer.atlassian.com/server/jira/platform/configuring-a-jira-cluster/)

Fixes #245

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -49,6 +49,8 @@ class jira (
   Hash $jira_config_properties                                      = {},
   Boolean $datacenter                                               = false,
   $shared_homedir                                                   = undef,
+  $ehcache_host                                                     = undef,
+  $ehcache_port                                                     = undef,
   # Database Settings
   Enum['postgresql','mysql','sqlserver','oracle','h2'] $db          = 'postgresql',
   $dbuser                                                           = 'jiraadm',
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -49,9 +49,9 @@ class jira (
   Hash $jira_config_properties                                      = {},
   Boolean $datacenter                                               = false,
   $shared_homedir                                                   = undef,
-  $ehcache_listener_host                                            = undef,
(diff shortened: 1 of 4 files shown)


manifests/init.pp
```diff
@@ -49,6 +49,9 @@
   Hash $jira_config_properties                                      = {},
   Boolean $datacenter                                               = false,
   $shared_homedir                                                   = undef,
+  $ehcache_listener_host                                            = undef,
```
bastelfreak: can you please add datatypes for the new parameters?

bastelfreak:
Hi @nrhtr, thanks for the PR. Can you please have a look at the used email address in the commits? It isn't associated with your github acount.

nrhtr:
Hi @bastelfreak. Will fix up the data types shortly. I'm using my work laptop which is why it has a different email configured from my personal Github account. I can commit with my personal email configured if you prefer?

bastelfreak:
Adding your corp mail address to github would also be fine. The used address just has to have a relation to the github account.

manifests/init.pp
bastelfreak: We can probably use this type? https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/types/port.pp
In this case please update the stdlib dependency in metadata.json to at least 4.25.0

nrhtr: Added data types and bumped stdlib dependency

bastelfreak:
Thanks!

PR #284, opened 2019-02-20
WIP: Support Oracle Service Names as well as SIDs

negast:

### DIFF
--- a/README.md
+++ b/README.md
@@ -271,6 +271,12 @@ Set the schema
 
 The Default value is 'public'
 
+##### `$oracle_SID`
+
+If set to false module will use an oracle service syntax rather than the SID notation. Changes jdbc:${db}:thin:@${dbserver}:${dbport_real}:${dbname} to jdbc:${db}:thin:@${dbserver}:${dbport_real}/${dbname}
+
+The Default value is true
+
 ##### `$poolsize`
 
 The connection pool size to the database, defaults to 20
--- a/README.md
+++ b/README.md
@@ -271,7 +271,7 @@ Set the schema
 
 The Default value is 'public'
 
-##### `$oracle_SID`
+##### `$oracle_sid`
 
 If set to false module will use an oracle service syntax rather than the SID notation. Changes jdbc:${db}:thin:@${dbserver}:${dbport_real}:${dbname} to jdbc:${db}:thin:@${dbserver}:${dbport_real}/${dbname}
 
--- a/README.md
+++ b/README.md
@@ -271,7 +271,7 @@ Set the schema
 
 The Default value is 'public'
 
-##### `$oracle_sid`
+##### `$oracle_use_sid`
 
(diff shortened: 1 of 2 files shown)


negast:
I added an additonal param called oracle_SID which is set to true by default. If set to true the old syntax is used else the service syntax is used. I added this param to the readme documentation.

manifests/init.pp
```diff
@@ -58,6 +58,7 @@
   $dbpassword                                                       = 'mypassword',
   $dbserver                                                         = 'localhost',
   $dbname                                                           = 'jira',
+  Boolean $oracle_sid                                                  = true,
```
bastelfreak: can you please update the alignment of `=` to match the others?

negast: Ok done

negast:
I don't get why the build is failing did I do something wrong here?

alexjfisher:
@negast Beaker tests can be a bit flakey. I restarted your build and it passed.

manifests/init.pp
```diff
@@ -231,13 +232,24 @@
     $dburl_real = $dburl
   }
   else {
-    $dburl_real = $db ? {
-      'postgresql' => "jdbc:${db}://${dbserver}:${dbport_real}/${dbname}",
-      'mysql'      => "jdbc:${db}://${dbserver}:${dbport_real}/${dbname}?useUnicode=true&amp;characterEncoding=UTF8&amp;sessionVariables=default_storage_engine=InnoDB",
-      'oracle'     => "jdbc:${db}:thin:@${dbserver}:${dbport_real}:${dbname}",
-      'sqlserver'  => "jdbc:jtds:${db}://${dbserver}:${dbport_real}/${dbname}",
-      'h2'         => "jdbc:h2:file:/${jira::homedir}/database/${dbname}",
+    if oracle_sid {
```
alexjfisher: How are these tests passing though?  Isn't this missing a leading `$`??

Also, is there a way of implementing this with less repetition?  Maybe by doing something like...
```puppet
$oracle_jdbc_separator = bool2str($oracle_sid, ':', '/')
```

README.md
```diff
@@ -271,6 +271,12 @@ Set the schema
 
 The Default value is 'public'
 
+##### `$oracle_sid`
```
alexjfisher: Could we change this parameter name?  `oracle_sid` doesn't sound like a Boolean option.  Maybe `oracle_use_sid`??

alexjfisher: Naming things is hard! ;)

manifests/init.pp
negast: Hmm yeah that does seem weird. I changed it to use the function as you suggested.

README.md
negast: I renamed it to the suggested param.

bastelfreak:
@negast can you take a look at the used email address in the commits? It isn't associated with your github account.

negast:
> @negast can you take a look at the used email address in the commits? It isn't associated with your github account.

I'm sorry where can I view this? I also changed my mail address recently for github.

bastelfreak:
The email address is correct now. You can see your avatar next to the commit, which means that you use a email address that is linked to a github account. Otherwise you can see the address with `git show $commitid`

oranenj:
I redid this in #365 , including credit to @negast. I'll close this PR

Issue #259, opened 2018-07-12
This logic does not allow Jira Service Desk to operate


cpepe:
https://github.com/voxpupuli/puppet-jira/blob/41eeb5c06ade0295fdfcea4760b2f2e239d44cd7/templates/server.xml.erb#L35

The Jasper listener needs to be disabled for JSD as well, due to the namely logic the product must be defined as "servicedesk" whereas the other two are "jira-core", and "jira-software"

cpepe:
Find the PR here: https://github.com/voxpupuli/puppet-jira/pull/260

bastelfreak:
thanks for reporting and fixing this @cpepe !

PR #288, opened 2019-03-06
Jira Service Desk doesn't start with JasperListener / need relaxedPath with correct version comp

thonixx:
Pull request of #260 (issue #259) wanted to prevent JasperListener to be
activated for Service Desk. Version comp condition can't be fulfilled as
Jira and Service Desk have different version numbers.

Therefore JasperListener gets added to the server.xml anyway with Service Desk. To get Service Desk working we need to adapt the condition to not match when we run Jira Service Desk.

Also relaxedPath options only needed from Jira version >v7.12.1 and Service Desk

(According to https://confluence.atlassian.com/jirakb/changing-server-xml-to-handle-requests-with-special-characters-958453799.html)

We need to have relaxedPathChars and relaxedQueryChars set according to
the FAQ page of Atlassian with Jira version greater than 7.12.1 and
apparently Service Desk.

Issue does not exist (yet), relates to issue #259 which doesn't work. Also relaxedPath should only be configured with the correct versions and for Service Desk.

### DIFF
--- a/templates/server.xml.erb
+++ b/templates/server.xml.erb
@@ -32,7 +32,7 @@
     <!--APR library loader. Documentation at /docs/apr.html -->
     <Listener className="org.apache.catalina.core.AprLifecycleListener" SSLEngine="on" />
     <Listener className="org.apache.catalina.mbeans.GlobalResourcesLifecycleListener"/>
-<%- if scope.function_versioncmp([@version, '6.4.14']) <= 0 && @product =~ /^jira|servicedesk/ -%>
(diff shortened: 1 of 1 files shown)


bastelfreak:
thanks for the fix @thonixx !

zogot:
Any chance to get this released? It's been a few months since this got out, but it would be valuable for those that are upgrading to version 8 of jira, of which the arguments are required.

bastelfreak:
hey @zogot, a new release depends on https://github.com/voxpupuli/puppet-jira/pull/298 and I currently don't have the time to look into this. Please let us know if you would like to help out. If so, you can have a look at the failed travis jobs and investigate why postgres isn't working.

PR #296, opened 2019-04-28
Add the ability to choose mysql db encoding

jordanconway:
As per https://confluence.atlassian.com/jirakb/emojis-and-special-characters-for-jira-with-mysql-952604493.html
You can use utf8mb4 encoding to allow the use of emoji in jira, this
patch allows users to set the character encoding in the db connection
string.

Signed-off-by: Jordan Conway <jconway@linuxfoundation.org>


jordanconway:
Went down the wrong path with this one.

PR #303, opened 2019-08-22
Adding ulimit option to setenv.sh environment file

rfreiberger:
As defined in the [Atlassian knowledge base article](https://confluence.atlassian.com/jirakb/health-check-open-files-limit-838555405.html), Jira can run out of open file resources leading stability issues. The fix is to add the following line for the `setenv.sh` file. 

```
ulimit -n 16384
```

The current `setenv.sh` file does not allow this option and since others may run into this problem, wanted to add this feature into the module. 

Adding the variable `ulimit` to allow defining ulimit settings in the `setenv.sh` file from the profile.

### DIFF
--- a/templates/setenv.sh.erb
+++ b/templates/setenv.sh.erb
@@ -3,7 +3,7 @@
 #
 JIRA_HOME="<%= scope.lookupvar('jira::homedir') %>"
 # ulimit options
-<% if scope.lookupvar('jira::ulimit') -%>
+<% if @jira::ulimit -%>
 ulimit -n "<%= scope.lookupvar('jira::ulimit') %>"
 <% end -%>
 #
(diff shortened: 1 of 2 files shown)


bastelfreak:
@rfreiberger can you please have a look at the failing travis jobs?

templates/setenv.sh.erb
```diff
@@ -2,7 +2,10 @@
 # One way to set the JIRA HOME path is here via this variable.  Simply uncomment it and set a valid path like /jira/home.  You can of course set it outside in the command terminal.  That will also work.
 #
 JIRA_HOME="<%= scope.lookupvar('jira::homedir') %>"
-
+# ulimit options
+<% if scope.lookupvar('jira::ulimit') -%>
```
bastelfreak: you can directly access the varaible instead of using `scope.lookupvar()`:
```suggestion
<% if @jira::ulimit -%>
```

rfreiberger:
I'm going to close this pull request, after checking the values on my test instance the settings as defined by Atlassian do not increase the values on their own. You need to apply the changes to the systemd limits as well. This was noted by Tommy in the Puppet Slack community channel. I'm not sure if it's valued to add this systemd step to the module since this may be a unique request but will be opening another pr with better testing in the near future. Thank you.

PR #316, opened 2020-04-28
Allow defining additional Tomcat connectors

antaflos:

Add a new, optional class parameter `jira::tomcat_additional_connectors`
whose value is a well-formed, complex hash as described by the new type
aliases `Jira::Tomcat_connectors` and `Jira::Tomcat_attributes`. Each
hash key represents a port number, and the key's value is a hash that
describes the connector's attributes and their values.

Useful for defining additional HTTP ports through which to access Jira.
A typical use case is when Jira and Confluence run behind a reverse
proxy and application links between Jira and Confluence must be set up.
Then it is often better to allow Jira and Confluence to communicate
directly instead of through the reverse proxy, but this requires a
Tomcat connector that is configured for direct access instead of reverse
proxy access (`proxyName`, `proxyPort`, etc). Described here:
https://confluence.atlassian.com/kb/how-to-create-an-unproxied-application-link-719095740.html

Contains spec tests, documentation and examples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -514,6 +514,23 @@ Password to access java keystore. Defaults to 'changeit'
 
 Defaults to 'JKS'. Valid options are 'JKS', 'PKCS12', 'JCEKS'.
 
+##### `$tomcat_additional_connectors`
+
+Well-formed, complex Hash where each key represents a port number and the key's
+value is a hash whose key/value pairs represent the attributes and their values
+that define the connector's behaviour. Default is `{}`.
+
+Use this parameter to specify arbitrary, additional connectors with arbitrary
+attributes. There are no defaults here, so you must take care to specify all
+attributes a connector requires to work in Jira. See below for examples.
+
+This is useful if you need to access your Jira instance directly through an
+additional HTTP port, e.g. one that is not configured for reverse proxy use.
+Atlassian describes use cases for this in
+https://confluence.atlassian.com/kb/how-to-create-an-unproxied-application-link-719095740.html
+and
+https://confluence.atlassian.com/kb/how-to-bypass-a-reverse-proxy-or-ssl-in-application-links-719095724.html
+
 #### Crowd single sign on parameters
 
 #### `enable_sso`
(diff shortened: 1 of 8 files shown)


antaflos:
Two out of four acceptance tests are failing in Travis but this is unrelated to my changes.

antaflos:
Trying to close and re-open to trigger Travis tests again.

antaflos:
The tests are fine now.

types/tomcat_attributes.pp
```diff
@@ -0,0 +1 @@
+type Jira::Tomcat_attributes = Hash[String[1], Scalar]
```
ghoneycutt, line 1: This does not seem to be used or documented anywhere. Is it meant to be here?

types/tomcat_connectors.pp
```diff
@@ -0,0 +1 @@
+type Jira::Tomcat_connectors = Hash[Stdlib::Port::Unprivileged, Jira::Tomcat_attributes]
```
ghoneycutt, line 1: Could you add unit tests for the types? Here's an example https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/spec/type_aliases/yes_no_spec.rb

You just have two arrays with things that match or don't.

README.md
```diff
@@ -611,6 +666,21 @@ jira::proxy:
   proxyName: 'jira.example.co.za'
   proxyPort: '443'
 jira::contextpath: '/jira'
+jira::tomcat_additional_connectors:
+  8181:
+    relaxedPathChars: '[]|'
+    relaxedQueryChars: '[]|{}^&#x5c;&#x60;&quot;&lt;&gt;'
```
bastelfreak, line 672: is this correct? The string looks a bit strange

antaflos: Looks strange, but see https://github.com/voxpupuli/puppet-jira/blob/master/templates/server.xml.erb#L54 on multiple lines, and https://confluence.atlassian.com/kb/how-to-create-an-unproxied-application-link-719095740.html and https://confluence.atlassian.com/jirakb/jira-server-startup-fails-with-tomcat-misconfigured-error-958453799.html for details on why this is necessary.

types/tomcat_connectors.pp
antaflos: Ok, will add tests for the types today or tomorrow, hopefully.

antaflos: Wow, that "today or tomorrow" was more than a week ago. Life and work got in the way but I haven't forgotten about this.

antaflos: I have now added spec tests for the type aliases and Travis is currently running the tests.

antaflos:
Some test suites are failing now because of lots of bizarre errors like this:

```
examples/jira_postgres_nginx_ssl_install.pp:35:manifest_whitespace_closing_bracket_before:ERROR:there should be no whitespace or a single newline before a closing bracket 
```

This is definitely not caused by my changes. I assume rubocop became much more strict in the last weeks?

antaflos:
Closing and reopening to trigger Travis, maybe the test failures unrelated to this change are now fixed.

antaflos:
The checks now pass so if the PR could be merged, that'd be great :)

bastelfreak:
thanks for the work!

PR #317, opened 2020-05-11
Adapting templates to state if 8.7.1 tarball

### LINKED ISSUE #300: Java 11 Compatibility
This parameter breaks compatibility with jdk11 -

https://github.com/voxpupuli/puppet-jira/blob/a40b36fbb8631d889b8691ff9e7aa51ad1d23ab5/templates/setenv.sh.erb#L50

Here is the exception -

```
Unrecognized VM option 'PrintGCDateStamps'
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.
NOTE: Picked up JDK_JAVA_OPTIONS:  --add-opens=java.base/java.lang=ALL-UNNAMED --add-opens=java.base/java.io=ALL-UNNAMED --add-opens=java.rmi/sun.rmi.transport=ALL-UNNAMED
```

Emptying this option allows Jira to start.

timdeluxe:
This adapts the templates for java-check, server.xml and setenv to the state of the 8.7.1 tarball, while keeping required modifications in place.

I thought about setting stuff depending on the installed java version, however not everyone installs java also via puppet and if ones does it, it must not necessarily be via the java puppet module.
Maybe we need an additional parameter to the module stating the java version?

Fixes #300


timdeluxe:
I see failing tests, but those are not related to my changes, but to a general issue: https://github.com/moby/moby/issues/38749
If you look at that issue, other puppet modules stopped the test on centos-7 completely or pinned it to a specific version. Maybe one of those solutions should be adapted to this module also?

timdeluxe:
I was too dumb updating my branch. I will close this and raise another PR. I am sorry.

PR #320, opened 2020-05-26
Modulesync 2.12.0, Fix Travis CI pipeline etc.

dhoppe:
Fixes #311


.sync.yml
```diff
@@ -3,4 +3,5 @@
   secure: "fH1bR1eBgA9tztMHtgdBSEOhNChC2YyoAczMzjUWgfDjinU1WOUPEprfV9upf6ZeQSTxAES36BBOrdxB6BVgdXlvLlCID+Lge1fTfi75ZX34q+eTadig/iZp3M0OveL2b5OmDv0cGbmMq8ywhgG+GPFuwPy7HF1Qjgul8UGNFOs="
   docker_sets:
     - set: centos7-64
-    - set: ubuntu1404-64
+spec/spec_helper_acceptance.rb:
+  unmanaged: true
```
ekohl, line 7: Isn't this the default?

manifests/params.pp
```diff
@@ -6,71 +6,49 @@
...
-            $service_provider = 'systemd'
-          } else {
-            $json_packages = [ 'rubygem-json', 'ruby-json' ]
-            $service_file_location = '/etc/init.d/jira'
-            $service_file_template = 'jira/jira.initscript.erb'
-            $service_lockfile = '/var/lock/jira'
-            $service_provider = 'debian'
-          }
-        }
-      }
```
ekohl, line 66: This drops support for Debian 7 and Ubuntu 14. That's fine, but it's probably good to do that in a separate PR first so it's included in the changelog.

metadata.json
```diff
@@ -14,15 +14,15 @@
...
-      "version_requirement": ">= 1.0.0 < 4.0.0"
+      "version_requirement": ">= 1.0.0 < 5.0.0"
     },
     {
       "name": "puppet/staging",
       "version_requirement": ">= 2.0.1 < 4.0.0"
     },
     {
       "name": "puppetlabs/stdlib",
-      "version_requirement": ">= 4.25.0 < 6.0.0"
```
ekohl, line 25: Should these 2 be done separate as well for the changelog?

spec/spec_helper_acceptance.rb
```diff
@@ -1,32 +1,12 @@
...
-  hosts.each do |host|
-    if host[:platform] =~ %r{el-7-x86_64} && host[:hypervisor] =~ %r{docker}
-      on(host, "sed -i '/nodocs/d' /etc/yum.conf")
-    end
-  end
+# This file is completely managed via modulesync
+require 'voxpupuli/acceptance/spec_helper_acceptance'
+
+configure_beaker do |_host|
+  install_module_from_forge('puppetlabs-apt', '>= 4.1.0 < 8.0.0') if fact('os.family') == 'Debian'
```
ekohl, line 5: I'd prefer to be explicit here
```suggestion
  install_module_from_forge_on(host, 'puppetlabs-apt', '>= 4.1.0 < 8.0.0') if fact_on(host, 'os.family') == 'Debian'
```

While we currently don't have multiple nodes, it's good to treat it as if we did

.sync.yml
dhoppe: Yes, but I wanted to make sure that this setting has not been forgotten. Otherwise someone could set it to false and break things.

manifests/params.pp
dhoppe: We already dropped support for Debian 7. I just picked a commit from @bastelfreak and did some cleanup:
* https://github.com/voxpupuli/puppet-jira/pull/320/commits/b216e8d77260b0aa21d6c724c94d887fc3bb2329

metadata.json
dhoppe: I just picked a commit from @alexjfisher and do not think that two separate commits would make a difference:
* https://github.com/voxpupuli/puppet-jira/pull/320/commits/78052cd8bc317c018c41d61590a00dc02b5dbcad

spec/spec_helper_acceptance.rb
dhoppe: Thank you very much for the review / suggestions.

.sync.yml
ekohl: You'd see that on the next modulesync and it'd fail. I wouldn't copy it.

spec/spec_helper_acceptance.rb
```diff
@@ -2,7 +2,7 @@
 require 'voxpupuli/acceptance/spec_helper_acceptance'
 
 configure_beaker do |_host|
-  install_module_from_forge('puppetlabs-apt', '>= 4.1.0 < 8.0.0') if fact('os.family') == 'Debian'
+  install_module_from_forge_on(host, 'puppetlabs-apt', '>= 4.1.0 < 8.0.0') if fact_on(host, 'os.family') == 'Debian'
```
ekohl, line 5: Note I suggested the on line, but you also need to change all others + rename `_host` to `host`.

.sync.yml
dhoppe: Alright. I hope you are right. ;)

spec/spec_helper_acceptance.rb
dhoppe: Fixed it.

PR #321, opened 2020-05-30
Update jira::facts to check for Puppet AIO agent

### LINKED ISSUE #151: jira::facts don't work with All-in-One Puppet 4.x agent
The AIO agent moved its ruby to /opt/puppetlabs/puppet/bin/ruby.  "PE" detection logic in jira::facts hasn't been adjusted for the AIO agent.  As a result, on nodes with the AIO agent, jira_facts.rb exists at both /etc/facter/facts.d/ and /etc/puppetlabs/facter/facts.d, and neither has a path to ruby that works, so both report Facter errors on agent runs.  I know that changing the shebang in jira_facts.rb to #!/opt/puppetlabs/puppet/bin/ruby fixes this.  I will try to put in a PR with updated logic for the AIO agent.

dpisano:

Updates `jira::facts` to check for the OSS version of the Puppet AIO agent. If the OSS Puppet AIO agent is installed it will set `facts.rb` to use the version of ruby that comes with the AIO agent.

Fixes #151

### DIFF
--- a/manifests/facts.pp
+++ b/manifests/facts.pp
@@ -32,6 +32,9 @@ class jira::facts (
   if $facts['puppetversion'] =~ /Puppet Enterprise/ {
     $ruby_bin = '/opt/puppet/bin/ruby'
     $dir = 'puppetlabs/'
+  } elsif $facts['aio_agent_version'] =~ String[1] {
+    $ruby_bin = '/opt/puppetlabs/puppet/bin/ruby'
+    $dir      = ''
   } else {
     $ruby_bin = '/usr/bin/env ruby'
     $dir = ''
--- a/manifests/facts.pp
+++ b/manifests/facts.pp
@@ -34,7 +34,7 @@ class jira::facts (
     $dir = 'puppetlabs/'
   } elsif $facts['aio_agent_version'] =~ String[1] {
     $ruby_bin = '/opt/puppetlabs/puppet/bin/ruby'
-    $dir      = ''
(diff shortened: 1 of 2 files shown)


manifests/facts.pp
```diff
@@ -32,6 +32,9 @@
   if $facts['puppetversion'] =~ /Puppet Enterprise/ {
     $ruby_bin = '/opt/puppet/bin/ruby'
     $dir = 'puppetlabs/'
+  } elsif $facts['aio_agent_version'] =~ String[1] {
+    $ruby_bin = '/opt/puppetlabs/puppet/bin/ruby'
+    $dir      = ''
```
ekohl, line 37: Shouldn't this also be:
```suggestion
    $dir      = 'puppetlabs/'
```

manifests/facts.pp
```diff
@@ -32,6 +32,9 @@
   if $facts['puppetversion'] =~ /Puppet Enterprise/ {
     $ruby_bin = '/opt/puppet/bin/ruby'
     $dir = 'puppetlabs/'
```
ekohl, line 34: I think this part is redundant now, assuming the other comment is correct.

dpisano: I don't have Puppet Enterprise to be able to test it to know for sure.

manifests/facts.pp
dpisano: Yes it should be. I will add it.

PR #324, opened 2020-07-22
Feature/templatize jvm extra args

bt-lemery:
Templatizing the JVM_EXTRA_ARGS parameter in setenv.sh to be able to accommodate custom settings.


jira.yaml
```diff
@@ -93,6 +93,7 @@ jira::jvm_optional: -XX:-HeapDumpOnOutOfMemoryError
 # the New and SR figures are purely optional
 # for heap dumps add -XX:-HeapDumpOnOutOfMemoryError
 # by default jira has 256m permgen which is a good setting to go with
+jira::jvm_extra: -XX:+PrintGCDateStamps
```
bastelfreak, line 96: Is it safe to not quote the strings? I'm never in yaml world :D

manifests/init.pp
```diff
@@ -90,6 +91,7 @@
   $jvm_xmx                                                          = '1024m',
   $jvm_permgen                                                      = '256m',
   $jvm_optional                                                     = '-XX:-HeapDumpOnOutOfMemoryError',
+  $jvm_extra                                                        = '-XX:+PrintGCDateStamps',
```
bastelfreak, line 94: We can enforce the datatype here as well:
```suggestion
  String $jvm_extra                                                 = '-XX:+PrintGCDateStamps',
```

templates/setenv.sh.erb
```diff
@@ -47,7 +47,7 @@ DISABLE_NOTIFICATIONS=" -Datlassian.mail.senddisabled=true -Datlassian.mail.fetc
 # This allows us to actually debug GC related issues by correlating timestamps
 # with other parts of the application logs.
 #-----------------------------------------------------------------------------------
-JVM_EXTRA_ARGS="-XX:+PrintGCDateStamps"
+JVM_EXTRA_ARGS="<%= scope.lookupvar('jira::jvm_extra') %>"
```
bastelfreak, line 50: instead of using the legacy scope.lookupvar, you should be able to directly access the variable with @jvm_extra

manifests/init.pp
bt-lemery: I've updated this.

templates/setenv.sh.erb
bt-lemery: scope.lookupvar is used everywhere else in these template, so was just attempting to follow current style.

oranenj:
Hi,

Thanks for the interest!

There's already something merged to master that achieves this. At any rate, I've overhauled the templates quite drastically recently, so fixing the conflicts in this one is going to be a huge pain. Therefore, I'll just close this PR.

PR #326, opened 2020-08-07
Make JVM_EXTRA_ARGS manageable for java 11 compatibility

diLLec:
With java-11-openjdk-11.0.7.10 the option "PrintGCTimeStamps" can't be used as a parameter on the JVM. Therefore this PR adds the capability to remove the parameter from setenv.sh.

Fixes the issue, that PrintGCTimeStamps is hardcoded in setenv.sh.erb template which results in a JVM startup error when java-11-openjdk is used.


diLLec:
also see: https://opennms.discourse.group/t/migration-of-jvm-settings-from-oracle-jdk-8-to-openjdk-11/365

manifests/init.pp
```diff
@@ -90,6 +90,7 @@
   $jvm_xmx                                                          = '1024m',
   $jvm_permgen                                                      = '256m',
   $jvm_optional                                                     = '-XX:-HeapDumpOnOutOfMemoryError',
+  $jvm_extra_args                                                   = '-XX:+PrintGCDateStamps'  # remove for openjdk 11
```
nambrosch, line 93: I think this could be changed to `'-XX:+PrintGCDateStamps -XX:+ExplicitGCInvokesConcurrent'` so jira doesn't complain about explicit garbage collection.  See -

https://confluence.atlassian.com/jirakb/health-check-explicit-garbage-collection-976778100.html

diLLec:
Hi @nambrosch - I've added the parameter to the default value and fixed another problem with the mysql_connector. I hope that we can both process them in one PR?

nambrosch:
> Hi @nambrosch - I've added the parameter to the default value and fixed another problem with the mysql_connector. I hope that we can both process them in one PR?

shortly after i wrote that comment, as a "solution" i started passing the flags `-XX:+ExplicitGCInvokesConcurrent -XX:+IgnoreUnrecognizedVMOptions -XX:+UseG1GC` through `java_opts` which makes jdk8, jdk11, confluence, and jira all happy.  I'm not sure what's better practice - leaving a commented-out line in your class or including the `-XX:+IgnoreUnrecognizedVMOptions` flag so all parameters can be included without consequence - just offering insight as to what worked for me.  I didn't see `-Xlog:::time,uptime,level,tags` in any documentation, so I'm not sure if it should be included here?

templates/setenv.sh.erb
```diff
@@ -47,7 +47,7 @@ DISABLE_NOTIFICATIONS=" -Datlassian.mail.senddisabled=true -Datlassian.mail.fetc
 # This allows us to actually debug GC related issues by correlating timestamps
 # with other parts of the application logs.
 #-----------------------------------------------------------------------------------
-JVM_EXTRA_ARGS="-XX:+PrintGCDateStamps"
+JVM_EXTRA_ARGS="<%= scope.lookupvar('jira::jvm_extra_args') %>"
```
bastelfreak, line 50: the variable should be accessible without the legacy `lookupvar()`
```suggestion
JVM_EXTRA_ARGS="<%= @jvm_extra_args %>"
```

igalic:
> I'm not sure what's better practice - leaving a commented-out line in your class or including the -XX:+IgnoreUnrecognizedVMOptions flag so all parameters can be included without consequence

@nambrosch, the consequence is that you're passing options, expecting them to have an effect, but they don't

i prefer a hard failure to hiding issues.

however, we shouldn't break what's worked so far

nambrosch:
> > I'm not sure what's better practice - leaving a commented-out line in your class or including the -XX:+IgnoreUnrecognizedVMOptions flag so all parameters can be included without consequence
> 
> @nambrosch, the consequence is that you're passing options, expecting them to have an effect, but they don't
> 
> i prefer a hard failure to hiding issues.
> 
> however, we shouldn't break what's worked so far

Normally I agree - it's a good tool for managing transitions between JDKs where puppet modules or other start scripts have hard-coded parameters that cause startup to fail.

manifests/init.pp
```diff
@@ -90,6 +91,13 @@
   $jvm_xmx                                                          = '1024m',
   $jvm_permgen                                                      = '256m',
   $jvm_optional                                                     = '-XX:-HeapDumpOnOutOfMemoryError',
+  $jvm_extra_args                                                   = '-XX:+PrintGCDateStamps -XX:+ExplicitGCInvokesConcurrent -XX:-OmitStackTraceInFastThrow -Djava.locale.providers=COMPAT',  # OpenJDK 1.8 / Oracle Java 8/11
```
bastelfreak, line 94: can you please add datatyppe for all new parameters?

templates/user.sh.erb
```diff
@@ -1,5 +1,5 @@
 # START INSTALLER MAGIC ! DO NOT EDIT !
-JIRA_USER="<%= scope.lookupvar('jira::user') %>" # user created by puppet
+JIRA_USER="<%= @user %>" # user created by puppet
```
bastelfreak, line 2: sweet, always :+1: to get rid of the legacy `scope.lookupvar()` function calls.

manifests/init.pp
```diff
@@ -86,12 +87,17 @@
   $pool_test_on_borrow                                              = false,
   # JVM Settings
   $javahome                                                         = undef,
-  $jvm_xms                                                          = '256m',
-  $jvm_xmx                                                          = '1024m',
-  $jvm_permgen                                                      = '256m',
-  $jvm_optional                                                     = '-XX:-HeapDumpOnOutOfMemoryError',
-  $java_opts                                                        = '',
-  $catalina_opts                                                    = '',
+  Enum['openjdk-11', 'oracle-jdk-1.8', 'custom'] $jvm_type          = 'custom',
```
igalic, line 90: given that this might just get more, it might be better to extract it in its own type

igalic (dismissed):
🙋🏻‍♀️

igalic:
i think i'd like @SimonHoenscheid's opinion on these changes, as he was working on a very similar patch

templates/setenv.sh.erb
```diff
@@ -37,87 +70,100 @@ DISABLE_NOTIFICATIONS=" -Datlassian.mail.senddisabled=true -Datlassian.mail.fetc
...
-PUSHED_DIR=`pwd`
-cd $LOGBASE
-cd $LOGTAIL
-LOGBASEABS=`pwd`
-cd $PUSHED_DIR
+PUSHED_DIR=$(pwd)
+cd ${LOGBASE}
+cd ${LOGTAIL}
+LOGBASEABS=$(pwd)
+cd ${PUSHED_DIR}
```
igalic, line 166: what's the point of this sequence?

igalic (commented):
🙋🏻‍♀️

SimonHoenscheid:
> i think i'd like @SimonHoenscheid's opinion on these changes, as he was working on a very similar patch

I see two things to make this change easy to maintain and understand:

1: Like I wrote [here](https://github.com/voxpupuli/puppet-jira/pull/329#discussion_r485168790)about supported versions and java dependencies, this would give us the option to depend on puppetlabs-java, therefore we can use these facts to make decisions more easy. 
2: switch a complete different template based on version or using two different blocks based on java version would make this easy to understand.

igalic:
thank you @SimonHoenscheid.

diLLec:
@bastelfreak @igalic @nambrosch I've have changed the code to whats been discussed in #329 and commented there. All checks are green now. Please review the PR :)

kenyon (changes requested):
Can you clean up the commit history? Thanks.

diLLec:
> 
> 
> Can you clean up the commit history? Thanks.

done and tests still green

kenyon (approved):

kenyon:
I just rebased on current master to get rid of the commits already in master.

manifests/init.pp
```diff
@@ -86,12 +87,21 @@
...
-  $jvm_xmx                                                          = '1024m',
-  $jvm_permgen                                                      = '256m',
-  $jvm_optional                                                     = '-XX:-HeapDumpOnOutOfMemoryError',
-  $java_opts                                                        = '',
-  $catalina_opts                                                    = '',
+  Jira::Jvm_types $jvm_type                                         = 'oracle-jdk-1.8',
+  String $jvm_xms                                                   = '256m',
+  String $jvm_xmx                                                   = '1024m',
+  String $jvm_permgen                                               = '256m',
+  Optional[String] $jvm_optional                                    = '',
```
bastelfreak, line 94: If the datatype is `Optional[]`, the value also needs to be `undef`. You cannot pass `undef` as a value to the jira class. Please switch the datatype to `String` or stick with `Optional` but change the default value from empty string (`''`) to `undef`.

kenyon: Good catch, fixed.

manifests/init.pp
```diff
@@ -71,6 +71,7 @@
   $mysql_connector_format                                           = 'tar.gz',
   Stdlib::Absolutepath $mysql_connector_install                     = '/opt/MySQL-connector',
   Stdlib::HTTPUrl $mysql_connector_url                              = 'https://dev.mysql.com/get/Downloads/Connector-J',
+  $mysql_connector_jar_suffix                                       = '-bin',
```
bastelfreak, line 74: Can you Add a datatype here as well? I guess `String[1]` works

types/jvm_types.pp
```diff
@@ -0,0 +1 @@
+type Jira::Jvm_types = Enum['openjdk-11', 'oracle-jdk-1.8', 'custom']
```
bastelfreak, line 1: can you add the puppet-strings style `# @summary ...` comment?

manifests/init.pp
kenyon: Done.

types/jvm_types.pp
kenyon: Done.

bastelfreak (approved):

templates/setenv.sh.erb
```diff
@@ -1,123 +1,169 @@
...
+ulimit -n $(( CURRENT_NOFILES_LIMIT > MIN_NOFILES_LIMIT  ? CURRENT_NOFILES_LIMIT : MIN_NOFILES_LIMIT ))
+
+PRGDIR=$(dirname "$0")
 cat "${PRGDIR}"/jirabanner.txt
 
 JIRA_HOME_MINUSD=""
-if [ "$JIRA_HOME" != "" ]; then
-    echo $JIRA_HOME | grep -q " "
-    if [ $? -eq 0 ]; then
+if [[ "${JIRA_HOME}" != "" ]]; then
```
ekohl, line 100: This bashism breaks on Ubuntu. Currently we don't run the acceptance tests on it, but in https://github.com/voxpupuli/puppet-jira/pull/334 I found out that it breaks.

PR #332, opened 2020-10-29
remove incorrect apostrophe in example

wolfaba:
this yaml code for jira::java_opts with incorrect apostrophe generates with the template `JAVA_OPTS="<%= scope.lookupvar('jira::java_opts') %> $JAVA_OPTS"` incorrect code 
```
JAVA_OPTS="-Datlassian.mail.senddisabled=true -Datlassian.mail.fetchdisabled=true -Datlassian.mail.popdisabled=true'
 $JAVA_OPTS"
```
which generates then error
```
/opt/jira/atlassian-jira-software-8.13.0-standalone/bin/catalina.sh: eval: line 517: unexpected EOF while looking for matching `''
/opt/jira/atlassian-jira-software-8.13.0-standalone/bin/catalina.sh: eval: line 519: syntax error: unexpected end of file
```
In fact, without apostrophe, it still generates linebreak in `JAVA_OPTS`, and then jira does not start. I will create ticket for this issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -644,7 +644,7 @@ jira::group:         'jira'
 jira::shell:         '/bin/bash'
 jira::dbserver:      'dbvip.example.co.za'
 jira::javahome:      '/opt/java'
-jira::java_opts: >
+jira::java_opts: >-
   -Dhttp.proxyHost=proxy.example.co.za
   -Dhttp.proxyPort=8080
   -Dhttps.proxyHost=proxy.example.co.za
--- a/README.md
+++ b/README.md
@@ -650,7 +650,7 @@ jira::java_opts: >
   -Dhttps.proxyHost=proxy.example.co.za
   -Dhttps.proxyPort=8080
   -Dhttp.nonProxyHosts=localhost\|127.0.0.1\|172.*.*.*\|10.*.*.*
-  -XX:+UseLargePages'
+  -XX:+UseLargePages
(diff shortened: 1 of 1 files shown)


kenyon (approved):

kenyon (approved):

kenyon:
It would be good if we had a test which fails before making this change.

wolfaba:
> It would be good if we had a test which fails before making this change.

@kenyon is this for me to make some fail-test or is it your internal note? Sorry, I'm not profi-puppet, so I have really no idea what to do :(

kenyon:
> 
> 
> > It would be good if we had a test which fails before making this change.
> 
> @kenyon is this for me to make some fail-test or is it your internal note? Sorry, I'm not profi-puppet, so I have really no idea what to do :(

@wolfaba this is a not for whoever might want to work on this 😉. To elaborate: if this is really a bug, the tests should have caught it by causing a failure. So if no such failing test exists, that test should be written. Then you can fix the bug and be sure you've fixed it and that it won't reappear without causing a test failure.

danifr:
I am also affected by this bug. As a workaround I defined all the JAVA_OPTS in a single line

kenyon:
I just realized that this is just a change to the docs, not the actual code, so disregard my note about tests. Merging. Thanks.

PR #335, opened 2020-11-26
Fix MySQL Connector directory permissions

### LINKED ISSUE #241: MySQL Connector is not installing with correct directory permissions

- Puppet: 5.3.5
- Ruby: 2.0.0p648
- Distribution: Puppet Enterprise
- Module version: 3.4.1

I installed JIRA with this configuration (after a successful MySQL installation):

```
  class { 'jira':
    version             => '7.8.1',
    installdir          => '/opt/atlassian-jira',
    db                  => 'mysql',
    dbtype              => 'mysql',
    dbname              => 'jiradb',
    dbuser              => 'jiradbuser',
    dbdriver            => 'com.mysql.jdbc.Driver',
    dbpassword          => 'xxx',
    dbserver            => 'localhost',
    dbport              => '3306',
    javahome            => '/usr/java/default',
    jvm_xmx             => '3584m',
    jvm_xms             => '3584m',
    download_url        => 'http://localrepo:6000/atlassian',
    mysql_connector_version => '5.1.46',
    mysql_connector_url => 'http://localrepo:6000/atlassian',
  }
```

## What are you seeing

Everything seems to install right in Puppet, but JIRA starts up and complains it can't find the MySQL connector JAR file. I can see it DID install though,  in `/opt/MySQL-connector`.

The problem appears to be that the JIRA module creates the permissions wrong for `/opt/MySQL-connector`. The directory is given permissions of `drwxr-x---` and is owned by root. This means user jira can't see inside the directory. When I manually change the permissions using `chmod a+rx /opt/MySQL-connector` and restart JIRA, then JIRA can see the database and things work fine.

Is there a workaround, or is it possible to get a new version of the JIRA module that sets the permissions correctly for /opt/MySQL-connector? I'd prefer it not be owned by user jira, as I'll be installing some other Atlassian tools on the same machine (and hopefully, they will use the same connector).

lgooch:
Set the directory permissions of the MySQL Connector directory so users other than root can use the connector.

Fixes #241

### DIFF
--- a/manifests/mysql_connector.pp
+++ b/manifests/mysql_connector.pp
@@ -15,6 +15,7 @@ class jira::mysql_connector (
       ensure => 'directory',
       owner  => root,
       group  => root,
+      mode   => '0755',
       before => Staging::File[$file],
     }
   }


oranenj:
Hi,

I redid this on top of current master where I reworked the mysql_connector manifest to use puppet/archive, so I'll close this PR.

PR #336, opened 2020-12-14
Drop EoL CentOS 6 support

bastelfreak:
CentOS 6 is not supported anymore. Our tests do not run on this
platform. They failed even before they reordered their mirrors, we just
didn't test on it. To unblock open PRs, this PR drops support for it.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -21,13 +21,6 @@ class jira::params {
         $service_file_mode     = '0644'
         $service_lockfile      = '/var/lock/subsys/jira'
         $service_provider      = 'systemd'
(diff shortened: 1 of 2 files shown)


kenyon (approved):

alexjfisher (commented):
Less than a week ago, we were in agreement about keeping code intact for a period of time (1 year?), unless there was a specific reason keeping it in was harmful to the module. This looks like a fairly trivial conditional.

ekohl:
In https://github.com/voxpupuli/puppet-jira/pull/344 I've proposed something that goes a step further and where it makes sense to hard remove all code.

kenyon:
Let's close this in favor of #344 then.

PR #337, opened 2021-01-19
Fixing that the suffix can be empty as well (needed for mysql-connector > 8)

diLLec:
The change is needed to specificy and use a mysql-connector version > 8. The problem with those connectors is, that there is no "-bin" suffix in the jarfile and therefore the suffix needs to be empty. As the current String will require the string to be at least 1 character long, a version 8 connector can not be configured.

```
class {'jira':
...
    mysql_connector_manage       => true,
    mysql_connector_version      => '8.0.23',
    mysql_connector_install      => '/opt/MySQL-connector',
    mysql_connector_url          => 'https://dev.mysql.com/get/Downloads/Connector-J',
    mysql_connector_jar_suffix   => '', # no '-bin'!
...
```
results in 
Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Resource Statement, Class[Jira]: parameter 'mysql_connector_jar_suffix' expects a String[1, default]

Enables the useage of mysql-connector version > 8

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -71,7 +71,6 @@ class jira (
   $mysql_connector_format                                           = 'tar.gz',
   Stdlib::Absolutepath $mysql_connector_install                     = '/opt/MySQL-connector',
   Stdlib::HTTPUrl $mysql_connector_url                              = 'https://dev.mysql.com/get/Downloads/Connector-J',
-  String[1] $mysql_connector_jar_suffix                             = '-bin',
   # Configure database settings if you are pooling connections
   $enable_connection_pooling                                        = false,
   $pool_min_size                                                    = 20,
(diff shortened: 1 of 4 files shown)


manifests/init.pp
```diff
@@ -71,7 +71,7 @@
   $mysql_connector_format                                           = 'tar.gz',
   Stdlib::Absolutepath $mysql_connector_install                     = '/opt/MySQL-connector',
   Stdlib::HTTPUrl $mysql_connector_url                              = 'https://dev.mysql.com/get/Downloads/Connector-J',
-  String[1] $mysql_connector_jar_suffix                             = '-bin',
+  String $mysql_connector_jar_suffix                                = '-bin',
```
bastelfreak, line 74: can you please add a test for this that ensures it works with an empty string? Otherwise people might readd `String[1]` in the future.

diLLec: Resolved and tests added - please see comment below

diLLec:
While working on the patch I've seen, that there already is version handlig code in the project (install.pp for example). Following that example I've put deleted the "suffix" variable and added logic that will change the filename scheme of the mysql connector jar based on the version.

@bastelfreak please review.

bastelfreak:
thanks for the update!

PR #340, opened 2021-02-14
acceptance tests: use the default version of postgresql

kenyon:
Version 9.4 doesn't exist in the postgresql.org repos anymore, which caused this test to fail. Simply using the default version that the puppetlabs-postgresql module wants works fine.

### DIFF
--- a/spec/acceptance/default_parameters_spec.rb
+++ b/spec/acceptance/default_parameters_spec.rb
@@ -7,11 +7,6 @@ describe 'jira postgresql' do
         distribution => 'jre',
       }
 
-      class { 'postgresql::globals':
-        manage_package_repo => true,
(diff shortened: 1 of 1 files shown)


spec/acceptance/default_parameters_spec.rb
```diff
@@ -9,7 +9,6 @@ class { 'java':
 
       class { 'postgresql::globals':
         manage_package_repo => true,
```
ekohl, line 11: Do we need to manage the package repo or is the OS version sufficient? Looking at the `postgresql::globals` code it then uses the OS default version so I'd think you can drop the entire `class { 'postgresql::globals': }` block.

ekohl (approved):

PR #347, opened 2021-03-01
Remove special code path for Puppet Enterprise

ekohl:
On Puppet Enterprise the aio_agent_version fact should also be set so it's redundant.

The json_packages code is made generic so it's executed on all non-AIO installs.

### DIFF
--- a/manifests/facts.pp
+++ b/manifests/facts.pp
@@ -26,35 +26,28 @@ class jira::facts (
   },
   # lint:endignore
 ) inherits jira::params {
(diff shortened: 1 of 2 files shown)


kenyon (approved):

kenyon:
Oh, looks like the tests didn't run for this. It looked good when I merged it, but the GitHub interface must have deceived me somehow.

ekohl:
Hmm, that is odd. Looks like it was never triggered and we don't have a required check.

PR #350, opened 2021-03-19
Add handling for X-Forwarded-For in access logs

diLLec:
Adding handling for X-Forwarded-For in access logs

If a proxy operates before JIRA, the access logs will only contain the IPs of the proxy instead of the IP of the user. With X-Forwarded-For the proxy can forward the users IP to the JIRA application server so that it can be logged correctly


kenyon:
Added README entry and data type.

kenyon (approved):

bastelfreak (approved):

PR #351, opened 2021-04-13
Restore $jira::java_opts for compatibility and remove $jira::jvm_*_additional from the API, as setting JDK type to custom allows full control anyway.

oranenj:
The _additional API is not useful at all; if people want to fully override their configuration, they can set their JDK type to custom

Test fixes missing for now, but I've tested this against an actual Jira instance

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -91,15 +91,11 @@ class jira (
   String $jvm_xmx                                                   = '1024m',
   String $jvm_permgen                                               = '256m',
   Optional[String] $jvm_optional                                    = undef,
-  Optional[String] $jvm_optional_additional                         = undef,
   Optional[String] $jvm_extra_args                                  = undef,
-  Optional[String] $jvm_extra_args_additional                       = undef,
   Optional[String] $jvm_gc_args                                     = undef,
-  Optional[String] $jvm_gc_args_additional                          = undef,
   Optional[String] $jvm_codecache_args                              = undef,
-  Optional[String] $jvm_codecache_args_additional                   = undef,
   Integer $jvm_nofiles_limit                                        = 16384,
-  String $java_opts                                                 = '',
+  Optional[String] $java_opts                                       = undef,
   String $catalina_opts                                             = '',
   # Misc Settings
   Stdlib::HTTPUrl $download_url                                     = 'https://product-downloads.atlassian.com/software/jira/downloads',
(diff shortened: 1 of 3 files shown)


manifests/init.pp
```diff
@@ -91,15 +91,11 @@
...
-  Optional[String] $jvm_optional_additional                         = undef,
   Optional[String] $jvm_extra_args                                  = undef,
-  Optional[String] $jvm_extra_args_additional                       = undef,
   Optional[String] $jvm_gc_args                                     = undef,
-  Optional[String] $jvm_gc_args_additional                          = undef,
   Optional[String] $jvm_codecache_args                              = undef,
-  Optional[String] $jvm_codecache_args_additional                   = undef,
   Integer $jvm_nofiles_limit                                        = 16384,
-  String $java_opts                                                 = '',
+  Optional[String] $java_opts                                       = undef,
```
bastelfreak, line 98: I think it doesn't make sense to pass `''` to the variable, so we could update it to this?
```suggestion
  Optional[String[1]] $java_opts                                       = undef,
```

oranenj: Empty `$java_opts` is perfectly valid currently, so that would be an API break. Since it gives a deprecation warning anyway, I think it's better to accept whatever configuration people have now.

bastelfreak: okay, fine by me

manifests/init.pp
```diff
@@ -173,6 +169,12 @@
     fail("\$shared_homedir must be set when \$datacenter is true")
   }
 
+  if $java_opts {
+    deprecation('jira::java_opts', 'jira::java_opts is no longer supported. Please use jira::jvm_extra_args')
+  }
+
+  $jvm_extra_args_real = "${java_opts} ${jvm_extra_args}"
```
ekohl, line 176:
```suggestion
  if $java_opts {
    deprecation('jira::java_opts', 'jira::java_opts is no longer supported. Please use jira::jvm_extra_args')
    $jvm_extra_args_real = "${java_opts} ${jvm_extra_args}"
  } else {
    $jvm_extra_args_real = $jvm_extra_args
  }
```

ekohl (approved):

oranenj:
squashed and reworded commit a bit

bastelfreak (approved):

PR #352, opened 2021-04-13
Refactor database configuration and make the main API typed

oranenj:
Refactors DB configuration to use a single template.

Changes PostgreSQL defaults a bit, but I'm pretty sure the default was ineffective anyway, so this is half of an API break...


oranenj:
Squashed and reworded a bit

bastelfreak (approved):

oranenj:
Needs another review, the changes are a bit more drastic now.

Basically, set defaults for ~everything to undef and omit them from dbconfig.xml unless specified by the user.

Also removes poolsize, I don't think it does anything, since it's not documented by Atlassian.

manifests/init.pp
```diff
@@ -99,50 +97,50 @@
...
   $deploy_module                                                    = 'archive',
-  $proxy_server                                                     = undef,
+  Optional[String] $proxy_server                                    = undef,
   Optional[Enum['none','http','https','ftp']] $proxy_type           = undef,
   # Manage service
-  $service_manage                                                   = true,
-  $service_ensure                                                   = running,
-  $service_enable                                                   = true,
+  Boolean $service_manage                                           = true,
+  Enum['running', 'stopped'] $service_ensure                        = 'running',
```
ekohl, line 108:
```suggestion
  Stdlib::Ensure::Service $service_ensure                           = 'running',
```

ekohl (approved):

oranenj:
I've also tested this on the Jira instance I have available and it doesn't break anything obvious, so I'll squash-merge.

PR #353, opened 2021-04-13
Allow specifying JDBC connection parameters with jira::connection_settings

oranenj:
Needs the dbconfig refactoring PR first. Adds an API to control JDBC connection parameters.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -62,6 +62,7 @@ class jira (
   Optional[String] $dbdriver                                        = undef,
   Optional[String] $dbtype                                          = undef,
   Optional[String] $dburl                                           = undef,
+  Optional[String] $connection_settings                             = undef,
   $dbschema                                                         = undef,
   # MySQL Connector Settings
   $mysql_connector_manage                                           = true,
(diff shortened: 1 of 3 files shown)


manifests/config.pp
```diff
@@ -19,25 +19,40 @@
...
-  if $jira::time_between_eviction_runs == undef {
-    $time_between_eviction_runs = $jira::db ? {
-      'postgresql' => '30000',
-      'mysql'      => '300000',
-      'oracle'     => '300000',
-      'sqlserver'  => '300000',
-      'h2'         => '5000',
-    }
+
+  $dbschema = if $jira::dbschema { $jira::dbschema } else { $dbschema_default }
```
ekohl, line 27: Is this because `pick` doesn't work with `undef`? Perhaps that deserves a note.

oranenj: I'll add it to the other PR; the first commit in this one is just because this depends on #352, so you can ignore that while reviewing.

bastelfreak (approved):

oranenj:
I renamed the parameter and added a spec test, let's see if I got it right.

oranenj:
I'll just merge this; if someone wants to bikeshed parameter naming, there's still time before release.

PR #354, opened 2021-04-14
Refactor most configuration to use EPP templates

oranenj:
This PR moves most configuration-related stuff from init.pp to config.pp, cleans up its dependencies and then converts most templates from ERB to EPP.

There are no API changes and I'm pretty sure the EPP templates result in identical output (at least they all stay the same on my Jira instance)


functions/sort_hash.pp
```diff
@@ -0,0 +1,4 @@
+function jira::sort_hash(Hash $input) {
```
ekohl, line 1: I think you can actually specify the output type and document it as a private API
```suggestion
# @summary Return a sorted hash
# @api private
function jira::sort_hash(Hash $input) >> Hash {
```

ekohl (approved):

bastelfreak (approved):

PR #357, opened 2021-04-15
Sync setenv.sh with upstream 8.13.5 (LTS)

oranenj:
This undeprecates $java_opts for backwards compatibility
and repurposes the other parameters to do what a custom $jvm_type
did previously.

It removes "-XX:HeapDumpOnOutOfMemoryError" because that's
not a default setting in the upstream setenv.sh

It also makes setting $jvm_permgen a no-op, since JIRA doesn't support
JDK older than 8 anymore and the option does nothing

### DIFF
--- a/README.md
+++ b/README.md
@@ -357,7 +357,7 @@ Defaults to `http://cdn.mysql.com/Downloads/Connector-J`
 
 ##### `$javahome`
 
-The JAVA_HOME directory, defaults to undef. This is a *required* parameter
+The `JAVA_HOME` directory, defaults to undef. This is a *required* parameter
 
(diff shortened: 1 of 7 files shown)


oranenj:
So I decided to remove the heapdump setting because it's not on by default and theoretically could leak some data from memory to  disk. I'm not sure why the module enabled it by default previously.

oranenj:
Though on second thought, I think less disruptive would be to just add a new parameter that's preferred over `jira::java_opts` if defined, and leave the old default in place...

oranenj:
The merge conflict was with a redundant test

bastelfreak (approved):

PR #359, opened 2021-04-15
Drop support for JIRA <8.0.0

oranenj:
Anything pre jira 8.0.0 is already EOL.

Drop support; legacy users can still use the old version of the module and use it to upgrade.

This only fixes manifests, I'll do templates separately. EDIT: ended up fixing templates anyway...

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -34,6 +34,12 @@ class jira::install {
     }
   }
 
+  file { $jira::homedir:
+    ensure => 'directory',
+    owner  => $jira::user,
+    group  => $jira::group,
+  }
+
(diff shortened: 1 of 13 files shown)


manifests/install.pp
```diff
@@ -82,29 +67,22 @@
     source          => "${jira::download_url}/${file}",
     creates         => "${jira::webappdir}/conf",
     cleanup         => true,
-    checksum_verify => $jira::checksum_verify,
+    checksum_verify => ($jira::checksum != undef),
```
bastelfreak, line 70: ha neat!

bastelfreak (approved):

PR #363, opened 2021-04-16
Provide option to install a java package directly via this module

oranenj:
This simplifies most configurations


manifests/init.pp
```diff
@@ -83,6 +83,7 @@
   Optional[Boolean] $pool_test_while_idle                           = undef,
   Optional[Boolean] $pool_test_on_borrow                            = undef,
   # JVM Settings
+  Optional[String] $java_package                                    = undef,
```
bastelfreak, line 86: please use this:
```suggestion
  Optional[String[1]] $java_package                                 = undef,
```

bastelfreak (approved):

PR #369, opened 2021-04-19
Bump default MySQL connector version to 8.0.23 and allow disabling the default HTTP connector

oranenj:


oranenj:
@ekohl Github's random review suggestion seems to want you to review most of my PRs :P I'm almost done though; as soon as the locale problem with CentOS 7 / Puppet 7 is done, the module ought to be pretty much good for release.

manifests/init.pp
```diff
@@ -143,6 +143,7 @@
   Hash $proxy                                                       = {},
   # Options for the AJP connector
   Hash $ajp                                                         = {},
+  Boolean $disable_default_connector                                = false,
```
ekohl, line 146: Negative booleans are generally a bad idea. How about
```suggestion
  Boolean $enable_default_connector                                = true,
```

Or just
```suggestion
  Boolean $default_connector                                = true,
```

ekohl (approved):

PR #372, opened 2021-04-20
Remove jira::facts and rewrite upgrade logic

oranenj:
There should be no need to depend on facts for upgrade.

We can bring the facts thing back if people really want it, but maybe after 5.0...


oranenj:
For now, I added an acceptance test that tries to test an upgrade scenario. I guess we might merge that into the "defaults" test so that there's less duplication.

oranenj:
woop, it's green. I'll rebase and restore the other OSes, and let's see...

bastelfreak (approved):

oranenj:
Rebased to squish a bugfix on Ubuntu 18 (systemctl path). I also changed managehome to false for the jira user; that makes more sense, since we create the directory anyway and it will contain jira data.

PR #375, opened 2021-04-22
Configure dbconfig.xml parameters explicitly to stop JIRA's healthcheck from complaining

oranenj:
Turns out JIRA works without these, but will complain, at least on
some versions. Also make the approach to variable naming a bit more
consistent.


oranenj:
Ref in atlassian doc. Apparently the part of documentation I was reading previously that talked about default values doesn't apply?

https://confluence.atlassian.com/jirakb/database-connection-settings-healthcheck-fails-in-jira-server-792636748.html

bastelfreak (approved):

oranenj:
Note that I chose to use `pick` in `jira::config´ instead of putting the defaults in `init.pp`'s parameter list because that way the rest of the code is consistent and all the related logic is in the same place.

PR #376, opened 2021-04-23
Release 5.0.1

oranenj:
Well, of course it was wrong. ._.


oranenj:
The bugfix sets some values (the eviction settings) that are not needed with non-MySQL databases, but if I understand the documentation correctly this time, JIRA also doesn't use them on non-MySQL databases, so just including them simplifies the code

PR #378, opened 2021-05-07
Allow managing installation and home directory modes and enforce installdir permissions by default

### LINKED ISSUE #377: JIRA will not start if the installation directory mode is missing o+x

- Puppet: v7.6.1
- Ruby: 2.0.0
- Distribution: CentOS 7.9
- Module version: 5.0.1

I have set the parameter $installdir and the module change the permissions to root not to user jira.

## What are you seeing
`$installdir => /opt/atlassian-jira`
after puppet agent was running on the machine i see that User jira have no permissions to execute the startup-script in this folder
```
root@jira:/opt# ll /opt/
total 0
drwxr-xr-x.  7 root root 146 28. Apr 10:35 .
dr-xr-xr-x. 20 root root 257 28. Apr 10:34 ..
drwxr-x---.  3 root root  94 28. Apr 10:35 atlassian-jira
```
systemd cant start jira and I see the following output in the `journalctl -xe`
```
Failed at step EXEC spawning /opt/atlassian-jira/atlassian-jira-software-8.13.6-standalone/bin/start-jira.sh: Permission denied
```
Maybe I think that this https://github.com/voxpupuli/puppet-jira/blob/master/manifests/install.pp#L45 is not working

oranenj:
Fixes #377 

This also adds a couple parameters to allow people to set their own permissions, at their own risk.


manifests/init.pp
```diff
@@ -270,6 +278,10 @@
   Boolean $manage_user                                              = true,
   String $user                                                      = 'jira',
   String $group                                                     = 'jira',
+  String $installdir_owner                                          = 'root',
```
bastelfreak, line 281: II think  empty string doesn't make sense?
```suggestion
  String[1] $installdir_owner                                       = 'root',
```

manifests/init.pp
```diff
@@ -270,6 +278,10 @@
...
+  String $installdir_group                                          = 'root',
```
bastelfreak, line 282:
```suggestion
  String[1] $installdir_group                                       = 'root',
```

manifests/init.pp
```diff
@@ -293,7 +305,7 @@
   Optional[String] $dburl                                           = undef,
   Optional[String] $connection_settings                             = undef,
   Boolean $oracle_use_sid                                           = true,
-  $dbschema                                                         = undef,
+  Optional[String] $dbschema                                        = undef,
```
bastelfreak, line 308:
```suggestion
  Optional[String[1]] $dbschema                                     = undef,
```

PR #381, opened 2021-05-18
Configure recommended JDBC connection-properties by default for PostgreSQL databases

### LINKED ISSUE #380: connection-settings parameter in dbconfig.xml template should be connection-properties instead
I guess this was a mental typo :-)

Will send a PR soon

oranenj:
For PostgreSQL only. Also fixes #380

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -115,8 +115,13 @@ class jira::config {
   $pool_test_while_idle = pick($jira::pool_test_while_idle, true)
   $pool_test_on_borrow = pick($jira::pool_test_on_borrow, false)
 
-  # This is just for consistency
-  $connection_settings = $jira::connection_settings
+  # JIRA will complain if these aren't set for PostgreSQL (will work fine though)
+  # https://confluence.atlassian.com/jirakb/connection-problems-to-postgresql-result-in-stuck-threads-in-jira-1047534091.html
+  if $jira::db == 'postgresql' {
+    $connection_settings = pick($jira::connection_settings, 'tcpKeepAlive=true;socketTimeout=240')
+  } else {
+    $connection_settings = $jira::connection_settings
+  }
 
   if $jira::db == 'mysql' {
     $validation_query_timeout = pick($jira::validation_query_timeout, 3)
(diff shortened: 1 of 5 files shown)


oranenj:
Reference: https://confluence.atlassian.com/jirakb/connection-problems-to-postgresql-result-in-stuck-threads-in-jira-1047534091.html

Installing the upgraded driver via the module would also be useful, but out of scope for this PR.

oranenj:
For most installations not configuring the values does not matter, but JIRA will complain about it on upgrades so let's do it.

manifests/config.pp
```diff
@@ -115,8 +115,13 @@
   $pool_test_while_idle = pick($jira::pool_test_while_idle, true)
   $pool_test_on_borrow = pick($jira::pool_test_on_borrow, false)
 
-  # This is just for consistency
-  $connection_settings = $jira::connection_settings
+  # JIRA will complain if these aren't set for PostgreSQL (will work fine though)
+  # https://confluence.atlassian.com/jirakb/connection-problems-to-postgresql-result-in-stuck-threads-in-jira-1047534091.html
+  if $jira::db == 'postgresql' {
```
bastelfreak, line 120: personally, I prefer a selector statement, but if/else works as well.

templates/dbconfig.xml.epp
```diff
@@ -51,7 +51,7 @@
     <time-between-eviction-runs-millis><%= $jira::config::time_between_eviction_runs %></time-between-eviction-runs-millis>
 <% } -%>
 <% if $jira::config::connection_settings { -%>
-    <connection-settings><%= $jira::config::connection_settings %></connection-settings>
+    <connection-properties><%= $jira::config::connection_settings %></connection-properties>
```
bastelfreak, line 54: is that a backwards incompatible change?

oranenj: No; the connection-settings option never did anything.

manifests/config.pp
oranenj: I prefer a simple if / else until there are more choices; I can never remember the selector syntax and it feels very clunky with only one option.

bastelfreak (approved):

oranenj:
What's up with Ubuntu failing? I did nothing that would affect those tests...

oranenj:
I'd like to merge this even though Ubuntu is failing. It seems there's just something in the `mysql` module that's not quite idempotent

oranenj:
Bah, attempting to fix the tests just makes things worse.

oranenj:
Since the only changes after approval were to make the MySQL tests run properly, I'll just merge this

PR #383, opened 2021-06-15
Extend the version check to the tailored jira servicedesk version

diLLec:
The current version check does not take the JIRA servicedesk package into account. This package has an other versioning schema - see here: https://www.atlassian.com/de/software/jira/service-management/download-archives

Currently, if the following parameters are choosen to init the jira class, the output shows

`JIRA versions older than 8.0.0 are no longer supported. Please use an older version of this module to upgrade first`

```puppet
class { 'jira':
    #
    # JIRA parameters
    version   => '4.17.0',
    product  => 'servicedesk',
...
```

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -412,7 +412,7 @@ class jira (
   Optional[Integer[0]] $poolsize                                    = undef,
   Optional[Boolean] $enable_connection_pooling                      = undef,
 ) {
-  // To maintain compatibility with previous behaviour, we check for not-servicedesk instead of specifying the 
+  # To maintain compatibility with previous behaviour, we check for not-servicedesk instead of specifying the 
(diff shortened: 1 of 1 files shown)


oranenj:
@diLLec the change breaks if someone is installing "jira-core". I confess I don't know if that's even a valid option anymore, but you need to add it to the enum or return it to a string and then fix the version comparison logic to take that into account.

oranenj (changes requested):
See above comment

diLLec:
Thanks @oranenj - I've added that option and the tests seam to be succeeding now.

oranenj:
I made the changes directly from Github, I think this should be good enough to fix the issue and not break API.

oranenj (approved):

bastelfreak (approved):

PR #384, opened 2021-06-18
Custom plugin installation; JNDI datasource

ThomasMinor:
* added the ability to install custom plugins via puppet
* addaed te ability to configure the usage of a JNDI datasource

n/a


spec/classes/jira_plugin_spec.rb
```diff
@@ -0,0 +1,85 @@
+require 'spec_helper.rb'
+
+describe 'jira' do
```
kenyon, line 3: Remove this level.

Also, this file should be named `spec/defines/jira_plugin_spec.rb`.

manifests/plugin.pp
```diff
@@ -0,0 +1,6 @@
+# @summary
```
kenyon, line 1: Unused?

templates/dbconfig.jndi.xml.epp
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<%# Some default are set in jira::config based on DB -%>
+<jira-database-config>
+  <name>defaultDS</name>
+  <delegator-name>default</delegator-name>
+  <database-type><%= $jira::config::dbtype %></database-type>
+<% if $jira::config::dbschema != undef { -%>
```
kenyon, line 8:
```suggestion
<% if $jira::config::dbschema { -%>
```

templates/dbconfig.jndi.xml.epp
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<%# Some default are set in jira::config based on DB -%>
```
kenyon, line 3:
```suggestion
<%# Some defaults are set in jira::config based on DB -%>
```

kenyon (changes requested):
Don't delete HISTORY.md, since it's included in the generated changelog.

manifests/init.pp
```diff
@@ -296,12 +296,14 @@
   Optional[Stdlib::Port] $ehcache_listener_port                     = undef,
   Optional[Stdlib::Port] $ehcache_object_port                       = undef,
   # Database Settings
-  Enum['postgresql','mysql','sqlserver','oracle','h2'] $db          = 'postgresql',
+  $use_jndi_ds                                                      = false,
+  $jndi_ds_name                                                     = 'JiraDS',
```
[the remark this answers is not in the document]
kenyon: Also, please add documentation comments for new parameters.

templates/dbconfig.jndi.xml.epp
ThomasMinor: I used the dbconfig.xml.epp as an example and wanted to keep it similar.

.rubocop.yml
```diff
@@ -1,3 +1,6 @@
 ---
 inherit_gem:
   voxpupuli-test: rubocop.yml
+
+AllCops:
```
bastelfreak, line 5: please don't modify this file, it's templated across all of our modules

manifests/config.pp
```diff
@@ -151,17 +153,21 @@
   # Configuration logic ends, resources begin:
 
   file { "${jira::webappdir}/bin/user.sh":
-    content => epp('jira/user.sh.epp'),
+    content => epp("${module_name}/user.sh.epp"),
```
bastelfreak, line 156: could you provide this change (and the other replacements of the hardcoded `jira` string) as a dedicated PR? That reduces the diff and makes the PR easier to review

templates/dbconfig.jndi.xml.epp
oranenj: There are some cases where you want to explicitly check for undef, but yeah, in this case I don't think it's necessary since an empty string is not false.

oranenj (changes requested):
I would like you to rebase the whole thing and split the changes into logical commits (order is up to you):
1) Cleanups to manifests
2) Functional changes (there's two here, plugin installation and the JNDI datasource, so if you can make two commits out of those that would be best)
3) The test changes + cleanup (preferably also separated to functional+cleanups, but it seems like a fairly involved change so if splitting is too much work just having them all in one is fine).

That way, reviewing the individual commits should give a clearer picture where actual user-impacting behaviour changes may occur.

Otherwise I don't see major issues.

manifests/config.pp
ThomasMinor: Ok, I'm fine with both strategies, so I suggest @oranenj an @bastelfreak agree on one or the other?

oranenj: Pick whichever approach you feel is best, I don't mind either way.

ThomasMinor:
I just had a bike accident so it‘ll take a vor till my next commit, sorry…

PR #385, opened 2021-07-06
use variable instead of literal in server.xml

dacron:
Using the `proxy` parameter when `tomcat_native_ssl` is true results in a string literal of "key" in server.xml in the SSL connector.

Pull request address that by using the previously declared `$key` variable in the EPP template.

n/a

### DIFF
--- a/templates/server.xml.epp
+++ b/templates/server.xml.epp
@@ -87,7 +87,7 @@
                     keystoreType="<%= $jira::tomcat_keystore_type %>"
 <%   if $jira::proxy { -%>
 <%     jira::sort_hash($jira::proxy).each |$key, $value| { -%>
-                    <%= key %> = '<%= $value %>'
+                    <%= $key %> = '<%= $value %>'
 <%     } -%>
 <%   } -%>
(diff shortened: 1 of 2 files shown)


dacron:
@bastelfreak this test failure doesn't look related to the change?

bastelfreak:
that's the kind of error you get if docker failed. I restarted the jobs.

templates/server.xml.epp
```diff
@@ -87,7 +87,7 @@
                     keystoreType="<%= $jira::tomcat_keystore_type %>"
 <%   if $jira::proxy { -%>
 <%     jira::sort_hash($jira::proxy).each |$key, $value| { -%>
-                    <%= key %> = '<%= $value %>'
+                    <%= $key %> = '<%= $value %>'
```
bastelfreak, line 90: that really looks like a bug that should have been catched by the unit tests. @dacron could you add a test for this that verifies the xml content?

dacron: I've added a test.

PR #389, opened 2021-08-13
Allow custom plugin installation and usage of JNDI database connections

ThomasMinor:
Added feature to install custom plugins and use JNDI database connections

n/a

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -58,6 +58,10 @@
 #   EHCache configuration for clustered mode
 # @param ehcache_object_port
 #   EHCache configuration for clustered mode
+# @param use_jndi_ds
+#   If true, the database will be configured as JNDI datasource in server.xml and then referenced in dbconfig.xml
+# @param jndi_ds_name
+#   Configures the JNDI datasource name
 # @param db
 #   The kind of database to use.
 # @param dbname
@@ -263,6 +267,15 @@
 #   undocumented SSO parameter
 # @param session_lastvalidation
 #   undocumented SSO parameter
+# @param plugins
+#   an array of hashes defining custom plugins to install
+#   a single plugin configuration will has the following form
+#     installation_name: this name wil be used to install the plugin within jira
+#     source: url of plugin to be fetched
+#     username: the username for authentification, if necessary
+#     password: the password for authentification, if necessary
+#     checksum: the checksum of the plugin, to determine the need for an upgrade
+#     checksumtype: the type of checksum used (none|md5|sha1|sha2|sha256|sha384|sha512). (default: none)
 # @param jvm_permgen
 #   Deprecated. Exists to notify users that they're trying to configure a parameter that has no effect
 # @param poolsize
@@ -296,6 +309,8 @@ class jira (
   Optional[Stdlib::Port] $ehcache_listener_port                     = undef,
   Optional[Stdlib::Port] $ehcache_object_port                       = undef,
   # Database Settings
+  Boolean $use_jndi_ds                                              = false,
+  String[1] $jndi_ds_name                                           = 'JiraDS',
   Enum['postgresql','mysql','sqlserver','oracle','h2'] $db          = 'postgresql',
   String $dbuser                                                    = 'jiraadm',
   String $dbpassword                                                = 'mypassword',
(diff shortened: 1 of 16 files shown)


manifests/init.pp
```diff
@@ -476,4 +492,25 @@
...
+        $target:
+          ensure        => 'absent',
+      }
+    } else {
+      archive {
+        $target:
+          ensure        => 'present',
+          source        => $plugin_data['source'],
+          username      => $plugin_data['username'],
+          password      => $plugin_data['password'],
```
root-expert, line 510: What happens if the plugin doesn't need username and password and the hash doesn't contain these keys? Maybe we should guard against it and default to `undef`?

Please add a test as well :smile:

spec/spec_helper.rb
```diff
@@ -16,3 +16,20 @@
...
+PATH_JIRA_DIR='/opt/jira'
+PATH_INSTALLATION_BASE = "#{PATH_JIRA_DIR}/#{PRODUCT_VERSION_STRING}-standalone"
+
+FILENAME_SETENV_SH = "#{PATH_INSTALLATION_BASE}/bin/setenv.sh"
+FILENAME_USER_SH = "#{PATH_INSTALLATION_BASE}/bin/user.sh"
+FILENAME_CHECK_JAVA_SH = "#{PATH_INSTALLATION_BASE}/bin/check-java.sh"
+FILENAME_SERVER_XML = "#{PATH_INSTALLATION_BASE}/conf/server.xml"
+FILENAME_DBCONFIG_XML = '/home/jira/dbconfig.xml'
+FILENAME_CLUSTER_PROPS='/home/jira/cluster.properties'
+FILENAME_JIRA_CONFIG_PROPS='/home/jira/jira-config.properties'
```
root-expert, line 35: `spec_helper.rb` is managed automatically by modulesync. 
if you want to modify it you need to change `.sync.yml`. Take a look at [this](https://github.com/voxpupuli/puppet-zabbix/blob/master/.sync.yml#L8) example.

Also maybe it's better to put these into a file and require it from the spec_helper.
If you need these attributes to be available in the acceptance tests as well you can place it in a file in the directory `spec/support/acceptance`, this will automatically loaded by the spec_acceptance_helper

root-expert (changes requested):
Hello! Thanks for your conribution!

Take a look at the inline comments I left. If you need any help you can contact us on the #voxpupuli channel on Libera.Chat IRC or on [Puppet's Slack](https://puppet.slack.com) (same channel)

ThomasMinor:
Hi there, does anyone have a hint how to eliminate the one failing test on CentOS 7? Maybe a good document about running the tests locally?

kenyon:
> Hi there, does anyone have a hint how to eliminate the one failing test on CentOS 7? Maybe a good document about running the tests locally?

Docs on running tests can be found here: https://github.com/voxpupuli/puppet-jira/blob/master/.github/CONTRIBUTING.md

root-expert:
Hey @ThomasMinor took a look at the output of the tests. Sadly we have seen this error before. The current workaround is to add [these 2 lines](https://github.com/voxpupuli/puppet-zabbix/blob/master/.github/workflows/ci.yml#L64-L65) in our pipeline.

ThomasMinor:
Hi @root-expert, I followed your advice to add the workaround and the tests succeeded - Thanks!
 I just have a few questions now...
- How did You identified the LANG settings as the cause of the problem from the logged messages?
- Isn't isn't it a bad idea to modify the ci.yml file due to the modulesync management?

bastelfreak (approved):

root-expert:
> * How did You identified the LANG settings as the cause of the problem from the logged messages?

The error was ` Error: /Stage[main]/Postgresql::Server::Reload/Exec[postgresql_reload]: invalid byte sequence in US-ASCII`

bastelfreak had deployed that workaround in the past for puppet-zabbix so it was known to me :smile: 
> * Isn't isn't it a bad idea to modify the ci.yml file due to the modulesync management?

Yes it will be overwritten, we manually add it back when we are modulesyncing.
There's some discussion [here](https://github.com/voxpupuli/modulesync_config/pull/708) regarding the issue btw.

root-expert (approved):
Not a user of the module but seems okay to me

root-expert:
Can you squash you commits before merging?

bastelfreak (approved):

PR #390, opened 2021-08-20
modulesync 4.2.0

bastelfreak:
modulesync 4.2.0


smortex (approved):
Rebased on top of master.

This fix the issue we had with a dependency.  The CentOS 7 error seems to be caused by an issue already reported here:
https://tickets.puppetlabs.com/browse/PUP-11196

PR #398, opened 2022-03-09
ulimit command does not apply to systemd services on all platforms

nambrosch:
Services started with systemd on CentOS 7, and possibly other platforms ignore ulimit unless the `LimitNOFILE` parameter is set in the systemd unit file.

https://bugzilla.redhat.com/show_bug.cgi?id=754285

### DIFF
--- a/templates/jira.service.epp
+++ b/templates/jira.service.epp
@@ -9,6 +9,7 @@ Environment="JAVA_HOME=<%= $jira::javahome %>"
 User=<%= $jira::user %>
 ExecStart=<%= $jira::webappdir %>/bin/start-jira.sh
 ExecStop=<%= $jira::webappdir %>/bin/stop-jira.sh
+LimitNOFILE=<%= $jira::jvm_nofiles_limit %>
 
 [Install]
 WantedBy=multi-user.target


jmcnatt:
Hey @root-expert or @bastelfreak - is it possible to get this fix in before the next release of the module?

This PR addresses the open file limits health check failure that could arise on systemd instances.  https://confluence.atlassian.com/jirakb/health-check-open-files-limit-838555405.html

root-expert:
Hello @nambrosch and @jmcnatt!

The patch seems okay but a unit test should be added before merging this.
I think the unit test can be added in [jira_service_spec](https://github.com/voxpupuli/puppet-jira/blob/master/spec/classes/jira_service_spec.rb) file.

nambrosch:
@root-expert Thanks for reviewing this, I don't think the service spec file needs to be updated since adding `LimitNOFILE` adds any new requirements, it's part of systemd, and it uses the value of a puppet variable, `$jira::jvm_nofiles_limit`, that already exists.

nambrosch:
@root-expert It's not obvious to me what should be added.  Any guidance that can be provided would be appreciated.

jmcnatt:
@nambrosch @root-expert would this be as simple as adding a content line in [./spec/classes/jira_service_spec.rb](https://github.com/voxpupuli/puppet-jira/blob/b3d860453dbbe7529e124c3ddb9c4f5a32ceb9b6/spec/classes/jira_service_spec.rb#L23)?

h-haaks:
@nambrosch could you please rebase this against voxpupuli:master?

h-haaks:
Hmm, on Ubuntu 18.04 and 20.04 this actually reduces the ulimit from 1048576 to 16384.
Is this a breaking change then?

PR #400, opened 2022-04-21
Update setenv.sh.epp if statement to work with service desk

techtino:
The version variable in puppet-jira can be either set to the jira software version or the servicedesk version. manifests/init.pp correctly differentiates but this template if statement does not, so the if statement does not apply to servicedesk > 4.11, missing out on an essential JVM parameter.

This PR adds in some and/ors to differentiate.

n/a


techtino:
Reopening after accidentally closing during re-base kerfuffle.

PR #403, opened 2023-04-14
change cluster nodeid to hostname

leightonium:
As fqdn could be longer than 60 characters in some cases, causing the cluster node to not start.

tests:
```
Finished in 3 minutes 30.9 seconds (files took 6.55 seconds to load)
1213 examples, 0 failures
```


h-haaks:
@leightonium could you please rebase your branch against our master?

PR #419, opened 2024-02-29
Change pool_max_wait type declaration

### LINKED ISSUE #418: Optional[Integer[-1]] $pool_max_wait breaks with latest puppet agent (7.29 or 8.5)

- Puppet: 7.29
- Ruby: all
- Distribution: all
- Module version: all (up to 6.0.0)

Upgrade puppet agent to the latest version (7.29 or 8.5)
and use this module

## What are you seeing
`
Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Resource Statement, The parameter '$pool_max_wait' must be a literal type, not a Puppet::Pops::Model::AccessExpression (file: /etc/puppetlabs/code/environments/production/external-modules/jira/manifests/init.pp, line: 333, column: 25) (file: /etc/puppetlabs/code/environments/production/manifests/site.pp, line: 3076) on node XXXX
`
successfull puppet run

This is the offending line
https://github.com/voxpupuli/puppet-jira/blob/b99c315018203a6728757c62538535470065e4d7/manifests/init.pp#L333

It is caused by a "bug" inside of puppet
https://github.com/puppetlabs/puppet/issues/9268

The current Integer min/max setting of -1 for pool_max_wait does not seem very useful anyhow, so perhaps better to remove the -1, and replace it with 0 like the other parameters

KoenDierckx:
This avoids an issue with latest puppet agent
But is not needed anyhow, as a minimal value of 0 already means wait indefinitely

Fixes #418

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -330,7 +330,7 @@ class jira (
   Stdlib::HTTPUrl $mysql_connector_url                              = 'https://dev.mysql.com/get/Downloads/Connector-J',
   Optional[Integer[0]] $pool_min_size                               = undef,
   Optional[Integer[0]] $pool_max_size                               = undef,
-  Optional[Integer[-1]] $pool_max_wait                              = undef,
+  Optional[Integer[0]] $pool_max_wait                               = undef,
   Optional[String] $validation_query                                = undef,
   Optional[Integer[0]] $validation_query_timeout                    = undef,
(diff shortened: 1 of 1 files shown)


kenyon:
From https://github.com/voxpupuli/puppet-jira/issues/418#issuecomment-1970759349 sounds like this should be closed.

PR #423, opened 2024-04-26
update puppet-systemd upper bound to 8.0.0

TheMeier:


TheMeier:
requires a new version for https://github.com/puppetlabs/puppetlabs-postgresql

PR #434, opened 2025-02-25
Fix breaking change regression of PR #429

diLLec:
Optimization for PR [Add "change database password" parameter - JIRA >10.3](https://github.com/voxpupuli/puppet-jira/pull/429)

* make change_dbpassword only be effective after version 10.3.0
* add tests


diLLec:
I really tried to add an acceptance test for JIRA 10.3.3 but I think we would need to limit those to certain modern OS versions like CentOS 9. Otherwise the prerequisites on MySQL and JAVA Version would not be able to be met. Unfortunately, I don't know how to limit the acceptance tests to a certain OS. If someone knows, please contact me.

templates/dbconfig.xml.epp
```diff
@@ -12,7 +12,7 @@
     <url><%= $jira::config::dburl %></url>
     <driver-class><%= $jira::config::dbdriver %></driver-class>
     <username><%= $jira::dbuser %></username>
-    <%- if $jira::config::change_dbpassword { -%>
+    <%- if versioncmp($jira::version, '10.3.0') < 0 or $jira::config::change_dbpassword { -%>
```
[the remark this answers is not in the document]
SimonHoenscheid: Mind the parameter order here, jira::version needs to be declared first

manifests/config.pp
```diff
@@ -48,7 +48,10 @@
     }
   }
 
-  $change_dbpassword = $jira::change_dbpassword
+  $change_dbpassword = versioncmp($jira::version, '10.3.0') ? {
+    /[0-1]/  => $change_dbpassword,
```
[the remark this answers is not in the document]
diLLec: All the parameters in the template templates/dbconfig.xml.epp are from jira::config. That's the reason I've done it that way. The tests are clean now. Please check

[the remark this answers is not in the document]
SimonHoenscheid: I don't see a reason to change the variable in the template. That is fine. 

But I would prefer it from a technical standpoint if the condition would move to init.pp and the parameter in config.pp would just inherit the value. For somebody looking at the init.pp it is not clear right now that a condition in config.pp might change the value. 

The rest of the PR looks good to me.

manifests/init.pp
```diff
@@ -484,6 +484,11 @@
     fail('You need to specify a value for javahome')
   }
 
+  $change_dbpassword = versioncmp($version, '10.3.0') ? {
```
SimonHoenscheid: Puppet does not allow reassignments.

There are two possible solutions:

1: do this in the parameter list, default to false.
2: use a variable like $real_change_dbpassword, to assign this to and use this one afterwards

manifests/config.pp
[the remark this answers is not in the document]
diLLec: I've added the code to init.pp but the puppet idiom of reassigned variables is not allowed makes it hard here. There are two options here: 

1. use another variable name like change_dbpassword_real  in init.pp and assign that in config.pp
```
  $change_dbpassword_real = versioncmp($version, '10.3.0') ? {
    -1      => true,
    default => $change_dbpassword,
  }
```
2. assign the variable to `jira::config` out of `init.pp`  like this
```
  $jira::config::change_dbpassword = versioncmp($version, '10.3.0') ? {
    -1      => true,
    default => $change_dbpassword,
  }
```

Do you see another option here? I've added a commit for version 2 but it did not work like that:
`Illegal attempt to assign to 'jira::config::change_dbpassword'. Cannot assign to variables in other namespace`

So the last commit implements option 1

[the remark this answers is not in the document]
SimonHoenscheid:
```
  $change_dbpassword = versioncmp($version, '10.3.0') ? {
    -1      => true,
    default => false,
},
```

would work, version needs to be declared first.

SimonHoenscheid:
@diLLec looks good to me. I asked for two other reviewers

SimonHoenscheid (approved):

SimonHoenscheid:
@diLLec can you sqash your commits so there is just one? I think most commits can be fixups.

`git rebase -i HEAD~13` should initiate the rebase

.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
```
bastelfreak, line 1: please don't commit your editor config

bastelfreak:
@diLLec thanks for the PR. As you mentioned, I think an acceptance test makes sense. https://github.com/voxpupuli/puppet-nodejs/blob/master/spec/acceptance/class_spec.rb#L38 has an example of limiting a test to specific operating systems. Does that help you?

.vscode/settings.json
SimonHoenscheid: should we add this to .gitigniore?

bastelfreak: I considered it. But our modulesync_config already purges it: https://github.com/voxpupuli/modulesync_config/blob/master/config_defaults.yml#L165-L166. So you could add it to the .gitignore, but it will be deleted when you run msync, and I think that would cause more confusion.

spec/classes/jira_config_spec.rb
```diff
@@ -91,6 +91,42 @@
             end
           end
 
+          context 'default params with java install and mysql > 10.3.0 and change_dbpassword=true' do
```
kenyon, line 94:
```suggestion
          context 'default params with java install, mysql, and jira > 10.3.0 and change_dbpassword=true' do
```

spec/classes/jira_config_spec.rb
```diff
@@ -91,6 +91,42 @@
...
+              }
+            end
+
+            it do
+              is_expected.to contain_file(FILENAME_DBCONFIG_XML).
+                with_content(%r{<password>test</password>})
+            end
+          end
+
+          context 'default params with java install and mysql > 10.3.0 and change_dbpassword=false' do
```
kenyon, line 112:
```suggestion
          context 'default params with java install, mysql, and jira > 10.3.0 and change_dbpassword=false' do
```

kenyon (commented):
We can squash commits on merge.

h-haaks:
The title of this pull request ends up in the change log.
Could you change the title to better address the fact that this fixes a regression introduced in https://github.com/voxpupuli/puppet-jira/pull/429 ?

h-haaks:
There is another regression as well.

when provisioning jira 10.3.x on a fresh server without specifying change_dbpassword = true, the password ends up being set to `{ATL_SECURED}` before the first server startup.

Solving this regression requires implementing some facts to figure out if jira is already installed or not.
It's probably best to address this in another issue/pull request.

diLLec:
After a day of testing and learning I finally managed to bring the acceptance test and the facts to work. Please review again.

diLLec:
Tests are fine now and commits are squashed. Please review and/or merge

vchepkov (changes requested):
All facts need to be confined, so they won't be failing on Windows, for example.
Also, do you need most of the facts? Normally module itself sets the desired state, here it feels like results are unpredictable?

diLLec:
> All facts need to be confined, so they won't be failing on Windows, for example. Also, do you need most of the facts? Normally module itself sets the desired state, here it feels like results are unpredictable?

Allright, I've confined the facts and made sure that the code that uses the fact can run without it beeing there. Most of the facts rely on each other. Sure, one can put that all in one fact but I found it more useful to single them out. 

The last sentence makes no sense to me. Could you elaborate more on what you mean?

vchepkov:
Happy to clarify! Besides the added database overhead of storing five values per node, these aren't truly 'facts' in the traditional sense. For example, if the desired status for the Jira service is 'active,' the 'jira_service_status' fact may become inaccurate after Puppet's execution if it wasn’t correct before. The same applies to the others, making these values unreliable as facts and unnecessarily increasing database utilization. A Deferred function might be a more efficient approach.

diLLec:
Thanks for pointing out your thoughts on this, but I think that sounds a bit acadamic. There are lots of facts that change while puppet runs - think of ethernet interfaces that are provisioned, or LVM volumes, or even the os.major which could change because of puppet changes. The Deferred function seams to be a way to get to the same result in a much more complex way than the current implementation.

diLLec:
OK, I've played a little with the deferred function and wonder how to test them inside of a acceptence test that uses no agent but puppet apply. 

If I puppet apply to following code, the result is as follows
```
$is_installed = Deferred('jira::is_installed', ['jira'])
notice($is_installed)
---
Notice: Scope(Class[main]): Deferred({'name' => 'jira::is_installed', 'arguments' => ['jira']})
```

If the function is used without Deferred, the result (as expected) is a correct Boolean:
```
$is_installed = jira::is_installed('jira')
notice($is_installed)
---
Notice: Scope(Class[main]): false
```

I would use the Deferred function like this in the init.pp

```
$is_installed = Deferred('jira::is_installed', [$jira::user])
```

Is something like this needed to make the acceptence tests run as expected?
```
  $is_insalled_deferred = Deferred('jira::is_installed', [$jira::user])
  $is_installed = $is_insalled_deferred =~ Deferred ? {
    true  => jira::is_installed($jira::user),
    false => $is_insalled_deferred
  }
```

vchepkov (commented):
I would pass a path to the config file instead of relying on user's home. 
manifest uses jira::homedir, imho, that should be used

lib/puppet/functions/jira/is_installed.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+# @summary Check if JIRA is already installed
+Puppet::Functions.create_function(:'jira::is_installed') do
```
bastelfreak, line 4: is there a reason why you used a function here that you defer? Usually such agent-side checks are done by a fact.

bastelfreak: Edit: Ah because you need to pass a user?

spec/acceptance/default_parameters_jira_10_spec.rb
```diff
@@ -0,0 +1,182 @@
+# frozen_string_literal: true
+
+require 'spec_helper_acceptance'
+
+def on_supported_os
+  (fact('os.family') == 'Debian' and fact('os.release.major') == '22.04') or (fact('os.family') == 'RedHat' and fact('os.release.major') >= '8')
```
bastelfreak, line 6: This excludes debian and future ubuntu versions. is that on purpose? And below you mentioned debian-11, but the test isn't running on that platform.

spec/acceptance/default_parameters_jira_10_spec.rb
```diff
@@ -0,0 +1,182 @@
...
+end
+
+prepare = <<-EOS
+  package {'diffutils':
+    ensure  => installed
+  }
+  file_line{'enable show_diff':
```
bastelfreak, line 13: was that actually required? I thought beaker uses that by default

spec/acceptance/default_parameters_jira_10_spec.rb
```diff
@@ -0,0 +1,182 @@
...
+    $manage_dnf_module = false
+    $autoremove_command = 'apt autoremove -y'
+  }
+  $jira_install_dir = '/opt/jira/'
+  $postgres_service = 'postgresql'
+  $jira_service = 'jira'
+EOS
+
+pp = <<-EOS
+  # The output of `systemctl status postgresql` is non ascii which
```
bastelfreak, line 47: I have some questions here. Why does that even break the exec resource? And is that a patch that should be added to the postgresql module instead?

spec/acceptance/default_parameters_jira_10_spec.rb
```diff
@@ -0,0 +1,182 @@
...
+  }
+  file_line{'enable show_diff':
+    path => '/etc/puppetlabs/puppet/puppet.conf',
+    line => 'show_diff = true'
+  }
+EOS
+
+pre = <<-EOS
+  if $facts['os']['family']  == 'RedHat' {
+    $java_package = 'java-17-openjdk'
```
bastelfreak, line 21: should those defaults be added to the actual puppet classes? Or at least be documented in the README.md?

diLLec:
> I would pass a path to the config file instead of relying on user's home. manifest uses jira::homedir, imho, that should be used

OK - added that

spec/acceptance/default_parameters_jira_10_spec.rb
diLLec: Yes - it is for testing purposes and to see more details. I've commented it out.

spec/acceptance/default_parameters_jira_10_spec.rb
diLLec: This I've copied from the existing default_parameters_spec.rb. I don't know if it is still needed.

spec/acceptance/default_parameters_jira_10_spec.rb
diLLec: Addred a comment on that in the readme.MD

spec/acceptance/default_parameters_jira_10_spec.rb
diLLec: debian11 and ubuntu2204 and RH8/9 are included now. Ubuntu 2004 only has postgresql 12 and therefore is not compatible with JIRA >10

diLLec:
From my point of view, this MR can be merged: all tests are OK and all the requested changes are included.

rwaffen:
okay, nice, just was looking for this change 😃 will merge and release

rwaffen (approved):

PR #439, opened 2025-02-27
Add Ubuntu 24.04

bastelfreak:


h-haaks:
Postgres 16 is installed on Ubuntu 24.04, but no Jira version support postgres > 15 ...
https://confluence.atlassian.com/adminjiraserver/supported-platforms-938846830.html

PR #440, opened 2025-02-27
Upgrade Jira test version to 9.12.18

SimonHoenscheid:
Raise Jira Test Version to 9.12.18

### DIFF
--- a/spec/acceptance/default_parameters_spec.rb
+++ b/spec/acceptance/default_parameters_spec.rb
@@ -6,13 +6,13 @@ describe 'jira postgresql' do
   it 'installs with defaults' do
     pp = <<-EOS
       $java_package = $facts['os']['family'] ? {
-        'RedHat' => 'java-11-openjdk-headless',
-        'Debian' => 'openjdk-11-jre-headless',
+        'RedHat' => 'java-17-openjdk-headless',
+        'Debian' => 'openjdk-17-jdk-headless',
       }
 
       $java_home = $facts['os']['family'] ? {
-        'RedHat' => '/usr/lib/jvm/jre-11-openjdk',
-        'Debian' => '/usr/lib/jvm/java-1.11.0-openjdk-amd64',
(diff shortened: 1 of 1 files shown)


spec/acceptance/default_parameters_spec.rb
```diff
@@ -40,17 +40,17 @@ class { 'jira':
...
       $java_home = $facts['os']['family'] ? {
-        'RedHat' => '/usr/lib/jvm/jre-11-openjdk',
-        'Debian' => '/usr/lib/jvm/java-1.11.0-openjdk-amd64',
+        'RedHat' => '/usr/lib/jvm/jre-17-openjdk',
+        'Debian' => '/usr/lib/jvm/java-17-openjdk-amd64',
       }
 
       class { 'jira':
-        version                  => '8.16.0',
+        version                  => '9.12.18',
```
bastelfreak, line 53: The version parameter in init.pp still defaults to 8. Should that be updated as well? Or do we want to keep the old version and test if the module can upgrade jira 8 to 9?

h-haaks: Jira 8.x, which has been eol a year now, does not support Java 17.
I guess most module users have moved to 9 by now, so I vote for bumping the default version :)

h-haaks: Hopefully that will solve the Debian 12/Ubuntu 24.04 issues as well.

spec/acceptance/default_parameters_spec.rb
```diff
@@ -6,13 +6,13 @@
   it 'installs with defaults' do
     pp = <<-EOS
       $java_package = $facts['os']['family'] ? {
-        'RedHat' => 'java-11-openjdk-headless',
-        'Debian' => 'openjdk-11-jre-headless',
+        'RedHat' => 'java-17-openjdk-headless',
+        'Debian' => 'openjdk-17-jdk-headless',
```
h-haaks, line 10: The module doesn't really support java17 that well ..

PR #454, opened 2025-06-23
Drop support for EOL Ubuntu 18.04 and Ubuntu 20.04

kenyon:
- #452
- **Drop support for EOL Ubuntu 18.04 and Ubuntu 20.04**

### DIFF
--- a/spec/acceptance/default_parameters_jira_10_spec.rb
+++ b/spec/acceptance/default_parameters_jira_10_spec.rb
@@ -5,7 +5,7 @@ require 'spec_helper_acceptance'
 def on_supported_os
   family = fact('os.family')
   major = fact('os.release.major')
-  (family == 'Debian' and (major >= '22.04' or (major >= '11' and major <= '20'))) or (family == 'RedHat' and major >= '8')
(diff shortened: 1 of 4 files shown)


spec/acceptance/default_parameters_jira_10_spec.rb
```diff
@@ -5,7 +5,7 @@
 def on_supported_os
   family = fact('os.family')
   major = fact('os.release.major')
-  (family == 'Debian' and (major >= '22.04' or (major >= '11' and major <= '20'))) or (family == 'RedHat' and major >= '8')
+  family == 'Debian' and (major >= '22.04' or (major >= '11' and major <= '20'))
```
jay7x, line 8: I'd maybe explicitly specify Ubuntu and Debian distros here instead of family + major.. but up to you

jay7x (approved):

spec/acceptance/default_parameters_jira_10_spec.rb
kenyon: > I'd maybe explicitly specify Ubuntu and Debian distros here instead of family + major.. but up to you

Yes, I think that would be better. I'll make another PR for that.

PR #464, opened 2025-08-13
Add support for Debian 12

smortex:


TheMeier:
```
 Error: Execution of '/usr/bin/apt-get -q -y -o DPkg::Options::=--force-confold install openjdk-11-jre-headless' returned 100: Reading package lists...
  Building dependency tree...
  Reading state information...
  Package openjdk-11-jre-headless is not available, but is referred to by another package.
  This may mean that the package is missing, has been obsoleted, or
  is only available from another source
  ```
D12 seems to come with java 17

